PR #1, opened 2020-11-12
Spring data neo4j impl

ao508:
This PR contains a basic working application with NATS pub/sub support and persisting objects into a neo4j graph db.

- CMO messaging library imported with jitpack
- Added main spring boot application class

TO-DO:
- We should expect every property of a SampleMetadata object to be updated when handling NEW_IGO_SAMPLE events
- At this time we can expect every property of a SampleMetadata object to require updates even if the sample already exists.
  - It is safe to assume that the IGO sample ID will be constant and unique
  - In the future we should be able to handle updates for a subset of properties
- Add timestamp to SampleMetadataEntity
- Identify properties that would trigger downstream processes to execute such as a field which would be used during the CMO sample ID generation process.
- Add CMO sample ID generation

Notes from Lisa
```
1.	IGO Sample ID trustworthy: Yes, an IGO sample ID never changes.
2.	Update field set: I think it will be easier for both of us to simply send a full set every time. But for your future use cases you might need to allow for subsets. If the sample receives an update post-pipeline, they might send a limited set.
```

### DIFF
--- a/persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
+++ b/persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
@@ -1,6 +1,6 @@
 package org.mskcc.cmo.metadb.persistence;
 
-import org.mskcc.cmo.messaging.model.PatientMetadata;
+import org.mskcc.cmo.shared.neo4j.PatientMetadata;
 
 import org.springframework.data.neo4j.annotation.Query;
 import org.springframework.data.neo4j.repository.Neo4jRepository;
@@ -11,16 +11,16 @@ import org.springframework.data.repository.query.Param;
  * @author ochoaa
  */
 public interface PatientMetadataRepository extends Neo4jRepository<PatientMetadata, Long> {
-    @Query("MATCH (pm:PatientMetadata) WHERE $investigatorPatientId = pm.investigatorPatientId RETURN pm")
+    @Query("MATCH (pm:cmo_metadb_patient_metadata) WHERE $investigatorPatientId = pm.investigatorPatientId RETURN pm")
     PatientMetadata findPatientByInvestigatorId(@Param("investigatorPatientId") String investigatorPatientId);
 
     @Query(
-            "MERGE (pm:PatientMetadata {investigatorPatientId: $patient.investigatorPatientId}) " +
+            "MERGE (pm:cmo_metadb_patient_metadata {investigatorPatientId: $patient.investigatorPatientId}) " +
                     "ON CREATE SET " +
-                        "pm.metaDbUuid = apoc.create.uuid(), pm.investigatorPatientId = $patient.investigatorPatientId " +
-                        "FOREACH (linkedPatient IN $patient.linkedPatientList | " +
-                            "MERGE (p:LinkedPatient {linkedPatientName: linkedPatient.linkedPatientName, linkedSystemName: linkedPatient.linkedSystemName}) " +
-                            "SET p = linkedPatient " +
+                        "pm.timestamp = timestamp(), pm.uuid = apoc.create.uuid(), pm.investigatorPatientId = $patient.investigatorPatientId " +
+                        "FOREACH (n_patient IN $patient.patientList | " +
+                            "MERGE (p {patientId: n_patient.patientId, idSource: n_patient.idSource}) " +
+                            "SET p = n_patient " +
                             "MERGE (p)-[:PX_TO_PX]->(pm) " +
                         ") " +
             "RETURN pm"
--- a/persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
+++ b/persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
@@ -1,0 +1,29 @@
+package org.mskcc.cmo.metadb.persistence;
+
+import org.mskcc.cmo.messaging.model.PatientMetadata;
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -0,0 +1,99 @@
...
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
+    <!-- java nats streaming -->
```
n1zea144, line 50: Does there have to be direct dependencies on nats outside of our messaging-java dependency?

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/SampleMetadataRepository.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * TODO: Follow up with LIMS/IGO folks.
+     *
+     *  - Will IGO sample ID be the same for a given sample even if it is updated in LIMS?
+     *  - In the event that we get an update for an existing sample in the metadb, will all of the properties for that sample be published or will only the updated properties be sent?
+     * @param sample
+     */
+    @Query(
+        "MERGE (sm:SampleMetadata {investigatorSampleId: $sample.investigatorSampleId}) " +
```
n1zea144, line 30: We should discuss this - I think that the cmo_metadb_sample_id (uuid) should be used as a primary key for metadata update.  After the node is created for the first time, we should be sharing the uuid to external systems for future reference to the node.

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.data.neo4j.annotation.Query;
+import org.springframework.data.neo4j.repository.Neo4jRepository;
+import org.springframework.data.repository.query.Param;
+
+/**
+ *
+ * @author ochoaa
+ */
+public interface PatientMetadataRepository extends Neo4jRepository<PatientMetadata, Long> {
+    @Query("MATCH (pm:PatientMetadata) WHERE $investigatorPatientId = pm.investigatorPatientId RETURN pm")
```
n1zea144, line 14: We should discuss this -  I'm not sure the utility of InvestigatorID - at the most fundamental level, I think that the cmo_metadb_patient_id (uuid) should be used as a primary key for metadata searching.   When a patient node is created in the graphdb, anytime we share patient metadata, we should be including the uuid.

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbMessagingApp.java
```diff
@@ -0,0 +1,78 @@
...
+    private PatientMetadataRepository patientMetadataRepository;
+
+    public static void main(String[] args) {
+        SpringApplication.run(MetadbMessagingApp.class, args);
+    }
+
+    @Override
+    public void run(String... args) throws Exception {
+        System.out.println("\n\nCleaning up graphdb...");
+        generalGraphDbRepository.deleteAllFromGraphDb();
```
n1zea144, line 37: Does this mean, every time we restart this app, the graph database gets wiped out?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleIntakeMessageServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private final String NEW_SAMPLE_INTAKE = "NEW_SAMPLE_INTAKE";
+    private final String NEW_SAMPLE_SUBMISSION = "NEW_SAMPLE_SUBMISSION";
+
+    private final Log LOG = LogFactory.getLog(SampleIntakeMessageServiceImpl.class);
+
+    @Autowired
+    private Gateway messagingGateway;
+
+    @Override
+    public void onSampleIntakeMessage() throws Exception {
```
n1zea144, line 28: From where and when does this get called?

n1zea144 (changes requested):

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/PatientMetadataRepository.java
ao508: this method was added as more of a placeholder/to help me manually test some things

pom.xml
ao508: i'll double check.. might have been left over from testing and/or the original version of the codebase I was working on

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbMessagingApp.java
ao508: yeah that's here just for testing right now

ao508: Sorry, should have added a note stating so

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleIntakeMessageServiceImpl.java
ao508: this is more temp code put in place where sample intake handling would eventually go

n1zea144 (approved):

PR #29, opened 2020-12-14
integrated neo4j connection with metadb running on dashi-dev

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/MetadbApp.java
```diff
@@ -28,7 +34,7 @@ public void run(String... args) throws Exception {
 
             // maybe create subscriber interface and pass handlers to messageGateway.subscribe?
             messagingGateway.subscribe(NEW_SAMPLE_INTAKE, SampleMetadataEntity.class,
-                    new SampleIntakeMessageConsumer(messagingGateway));
+                    new SampleIntakeMessageConsumer(messagingGateway, sampleMetadataRepository));
```
n1zea144, line 37: Unless there is another "consumer" in this class that requires a sampleMetadataRepository, why not autowire it directly in SampleIntakeMessageConsumer?

service/src/main/java/org/mskcc/cmo/metadb/service/consumer/SampleIntakeMessageConsumer.java
```diff
@@ -1,35 +1,71 @@
...
+        return pMetadata;
+    }
+    
+    private boolean addSampleMetadata(Object message) {
         try {
-            LOG.info("Publishing on topic: cmo.new-sample-submission\n");
-            messagingGateway.publish("cmo.new-sample-submission",
-                    "\"message from igo.new-sample-intake handler to cmo.new-sample-submission\"");
+            LOG.info("*** Persisting to NEO4j ***\n");
+            Gson gson = new Gson();
```
n1zea144, line 59: I would separate concerns here - that is, addSampleMetadata should take a SampleMetadataEntity, not a message.  gson.fromJson should be called in onMessage and the exception caught in onMessage().

service/src/main/java/org/mskcc/cmo/metadb/service/consumer/SampleIntakeMessageConsumer.java
```diff
@@ -1,35 +1,71 @@
...
         this.messagingGateway = messagingGateway;
+        this.sampleMetadataRepository = sampleMetadataRepository;
     }
 
     @Override
     public void onMessage(Object message) {
+        
         LOG.info("*** Message received on igo.new-sample-intake topic ***\n");
         LOG.info(message);
+        if (addSampleMetadata(message)) {
```
n1zea144, line 37: See my comments in addSampleMetadata().  I would call gson.fromJson() here and catch any exceptions before call addSampleMetadata().

service/src/main/java/org/mskcc/cmo/metadb/service/consumer/SampleIntakeMessageConsumer.java
```diff
@@ -1,35 +1,71 @@
...
+                LOG.info("Publishing on topic: cmo.new-sample-submission\n");
+                messagingGateway.publish("cmo.new-sample-submission",
+                        "\"message from igo.new-sample-intake handler to cmo.new-sample-submission\"");
+            } catch (Exception e) {
+                LOG.error("Error encountered during publishing on topic: cmo.new-sample-submission", e);
+            }
+        } 
+    }
+    
+    private PatientMetadata mockPatientMetadata(String patientId) {
```
n1zea144, line 48: Do we need this routine?  Can't our test publisher provide the necessary patient metadata?

n1zea144 (changes requested):

service/src/main/java/org/mskcc/cmo/metadb/service/consumer/SampleIntakeMessageConsumer.java
divyamadala30: This is temporary until we are able to provide patient data and update the neo4j query

service/src/main/java/org/mskcc/cmo/metadb/MetadbApp.java
ao508: Tried autowiring the components to the consumer directly but we were running into issues where the components were null. 

This is temporary for now until we resolve the autowiring issue.

service/src/main/java/org/mskcc/cmo/metadb/service/consumer/SampleIntakeMessageConsumer.java
```diff
@@ -1,35 +1,71 @@
...
-@Component
+@Service
+@ComponentScan(basePackages = "org.mskcc.cmo.metadb.persistence")
 public class SampleIntakeMessageConsumer implements MessageConsumer {
     private Gateway messagingGateway;
     private final Log LOG = LogFactory.getLog(SampleIntakeMessageConsumer.class);
+    private SampleMetadataRepository sampleMetadataRepository;
 
-    public SampleIntakeMessageConsumer(Gateway messagingGateway) {
+    public SampleIntakeMessageConsumer(Gateway messagingGateway,
```
n1zea144, line 26: I would rename this class to IGOSampleIntakeMessageConsumer (in the future we may intake samples from some other source).

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/SampleMetadataRepository.java
```diff
@@ -35,5 +38,5 @@
         + "MERGE (pm)-[:PX_TO_SP]->(sm)"
         + "RETURN sm"
     )
-    void saveSampleMetadata(@Param("sample") SampleMetadataEntity sample);
+    void saveSampleMetadata(@Param("sample") SampleMetadataEntity sample);
```
n1zea144, line 41: For the purposes of concurrency, we will need to have this method return the SampleMetadataEntity that was just persisted (including the uuid that was generated).

PR #73, opened 2021-01-20
CmoRequestEntity now extends IgoRequest

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/CmoRequestEntity.java
```diff
@@ -11,7 +12,7 @@
  * @author ochoaa
  */
 @NodeEntity(label = "cmo_metadb_request")
-public class CmoRequestEntity {
+public class CmoRequestEntity extends IgoRequest {
     @Id
     private String requestId;
```
ao508, line 17: requestId is now coming from IgoRequest but I think the graphdb will need still need an `@Id` annotation on one of the fields in the class. Let's use the standard `private Long id`, I think there are some examples in this repo where the node entity has that type of `@Id` set

ao508 (changes requested):
one minor change requested!

related to the requestId in the cmo request entity, we can remove the getters/setters for it since it's coming from the `IgoRequest` model

PR #74, opened 2021-01-21
CmoRequestEntity now extends IgoRequest

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/CmoRequestEntity.java
```diff
@@ -11,9 +13,9 @@
  * @author ochoaa
  */
 @NodeEntity(label = "cmo_metadb_request")
-public class CmoRequestEntity {
-    @Id
-    private String requestId;
+public class CmoRequestEntity extends IgoRequest {
+    @Id@GeneratedValue
```
ao508, line 17: can you add a space between these annotations?

ao508 (changes requested):
one more minor fix

ao508 (approved):
looks good! let's rebase and then I'll merge!

PR #84, opened 2021-01-27
Persist project, patient, and sampleManifestJson to Neo4J

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/CmoRequestEntity.java
```diff
@@ -18,6 +18,8 @@
     private Long id;
     @Relationship(type = "REQUEST_TO_SP", direction = Relationship.OUTGOING)
     private List<SampleManifestEntity> sampleManifestList;
+    @Relationship(type = "REQUEST_TO_SP", direction = Relationship.OUTGOING)
+    private List<NormalSampleManifestEntity> normalSampleManifestList;
```
n1zea144, line 22: Why you have to make a distinction between NormalSampleManifestEntity and SampleManifestEntity?

PR #85, opened 2021-01-28
Removed PX_TO_NORMAL edge and persist SampleManifestJson

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
```diff
@@ -19,6 +19,7 @@
     @Id @GeneratedValue(strategy = UuidStrategy.class)
     @Convert(UuidStringConverter.class)
     private UUID uuid;
+    private Boolean isTumor;
```
n1zea144, line 22: Is this required if we have tumorOrNormal?

model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
```diff
@@ -135,6 +136,14 @@ public UUID getUuid() {
     public void setUuid(UUID uuid) {
         this.uuid = uuid;
     }
+    
+    public Boolean getIsTumor() {
```
n1zea144, line 140: Should this just be (tumorOrNormal.equals("TUMOR") ?  I don't suggest we hardcode the string - we should use an Enum and make sure we know the vocabulary for this field.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/CmoRequestServiceImpl.java
```diff
@@ -19,17 +19,18 @@
...
-    public void saveRequest(CmoRequestEntity request) {
+    public void saveRequest(CmoRequestEntity request) throws Exception {
         //TODO: NEED TO PERSIST THE PROJECT DATA AS WELL
         CmoRequestEntity savedRequest = getCmoRequest(request.getRequestId());
         if (savedRequest == null) {
             if (request.getSampleManifestList() != null) {
                 for (SampleManifestEntity s: request.getSampleManifestList()) {
-                    sampleManifestRepository.save(s);
+                    sampleService.saveSampleManifest(s);
+                    //sampleService.findMatchedNormalSample(s);
```
n1zea144, line 33: Lets remove this comment.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -20,28 +22,63 @@
...
+        sampleManifestEntity.addSample(investigatorId);
+
+        SampleManifestJsonEntity sampleJson = new SampleManifestJsonEntity();
+        Timestamp timestamp = new Timestamp(System.currentTimeMillis());
+        sampleJson.setCreationDate(String.valueOf(timestamp.getTime()));
+        sampleJson.setSampleManifestJson(sampleManifestEntity.toString());
+        sampleManifestEntity.setSampleManifestJsonEntity(sampleJson);
+        
+        //TODO: check SampleManifestEntity mapping
+        if (sampleManifestEntity.getTumorOrNormal().equals("Tumor")) {
```
n1zea144, line 70: I would introduce an Enum in the SampleManifestEntity class which captures Tumor and Normal elements.

ao508 (approved):
there's some trailing whitespaces and we need to rebase before merging but the code itself looks good to me

PR #95, opened 2021-02-03
GraphDB refactoring

ao508:
Resolves issues:
- #92
- #89

### DIFF
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/PatientAlias.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/PatientAlias.java
@@ -12,19 +12,19 @@ import org.neo4j.ogm.annotation.Relationship;
  * @author ochoaa
  */
 
-@NodeEntity(label = "patient")
-public class Patient implements Serializable {
+@NodeEntity(label = "patient_alias")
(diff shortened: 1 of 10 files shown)


model/src/main/java/org/mskcc/cmo/metadb/model/PatientAlias.java
```diff
@@ -12,19 +12,19 @@
  * @author ochoaa
  */
 
-@NodeEntity(label = "patient")
-public class Patient implements Serializable {
+@NodeEntity(label = "patient_alias")
```
ao508, line 15: remove the label, just keep `@NodeEntity` annotations for everything

model/src/main/java/org/mskcc/cmo/metadb/model/SampleAlias.java
```diff
@@ -11,24 +11,24 @@
  * @author ochoaa
  */
 
-@NodeEntity(label = "sample")
-public class Sample implements Serializable {
+@NodeEntity(label = "sample_alias")
```
ao508, line 14: same comment here regarding `@NodeEntity` annotation

PR #107, opened 2021-02-04
added sampleClass to MetaDbSample

divyamadala30:


persistence/src/main/java/org/mskcc/cmo/metadb/persistence/MetaDbSampleRepository.java
```diff
@@ -35,7 +35,7 @@
     @Query("MATCH (s: MetaDbSample {uuid: $sampleManifestEntity.uuid})"
             + "MATCH (s)<-[:HAS_SAMPLE]-(p: MetaDbPatient)"
             + "MATCH (n)<-[:HAS_SAMPLE]-(p) "
-            + "WHERE n.tumorOrNormal = 'Normal'"
+            + "WHERE n.sampleClass = 'Normal'"
```
ao508, line 38: I'm curious whether we can make this query case insensitive or if it already ignores case when making string comparisons by default

ao508: cases:
1. search for 'normal' (all lowercase), does it return the same results as searching for 'Normal'?

PR #145, opened 2021-02-18
added controller for cmo-new-request and minor changes to repositories

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/service/impl/CmoRequestServiceImpl.java
```diff
@@ -59,6 +59,9 @@ public void saveRequest(MetaDbRequest request) throws Exception {
     @Override
     public Map<String, Object> getMetaDbRequest(String requestId) throws Exception {
         MetaDbRequest metaDbRequest = metaDbRequestRepository.findByRequestId(requestId);
+        if (metaDbRequest == null) {
+            return null;
```
ao508, line 63: maybe log that the returned request is null here too

web/.gitignore
```diff
@@ -0,0 +1,37 @@
+HELP.md
```
ao508, line 1: let's clean this block up a bit and replace with this

```
.mvn
**/target
mvnw*
```

web/.gitignore
```diff
@@ -0,0 +1,37 @@
...
+*.iml
+*.ipr
+
+### NetBeans ###
+/nbproject/private/
+/nbbuild/
+/dist/
+/nbdist/
+/.nb-gradle/
+build/
```
ao508, line 32: and these 3 lines replace with 

```
**/build
```

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,66 @@
...
+    @Autowired
+    public RequestController(CmoRequestService requestService) {
+        this.requestService = requestService;
+    }
+    
+    /**
+     * fetchRequestGET
+     * @param requestId
+     */
+    @ApiOperation(value = "Retrieve NewCmoRequest",
```
ao508, line 36: "NewCmoRequest" --> "MetaDB Request"

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,66 @@
...
+        nickname = "fetchRequestGET")
+    @RequestMapping(value = "/request/requestId/{requestId}",
+        method = RequestMethod.GET,
+        produces = "application/json")
+    @ResponseBody
+    public Map<String, Object> fetchRequestGET(@ApiParam(value = "requestId",
```
ao508, line 42: "fetchRequestGET" --> "fetchMetaDbRequestGET"

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * fetchRequestListPOST
+     * @param requestIds
+     */
+    @ApiOperation(value = "Retrieves list of MetaDbRequest from a list of RequestId.", 
+        nickname = "fetchRequestListPOST")
+    @RequestMapping(value = "request/requestId",
+        method = RequestMethod.POST,
+        produces = "application/json")
+    public List<Map<String, Object>> fetchRequestListPOST(@ApiParam(value =
```
ao508, line 57: "fetchRequestListPOST" --> "fetchMetaDbRequestPOST"

ao508 (changes requested):
some minor changes requested but looks good!

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,66 @@
...
+        this.requestService = requestService;
+    }
+    
+    /**
+     * fetchRequestGET
+     * @param requestId
+     */
+    @ApiOperation(value = "Retrieve NewCmoRequest",
+        nickname = "fetchRequestGET")
+    @RequestMapping(value = "/request/requestId/{requestId}",
```
ao508, line 38: actually I just realized that this mapping seems a bit weird. 

It should be value = "/request/{requestId}"

and the post should be value = "/request"

ao508 (changes requested):

service/pom.xml
```diff
@@ -11,12 +11,24 @@
...
+  <properties>
+    <slf4j.version>1.7.30</slf4j.version>
+  </properties>
   <dependencies>
     <dependency>
       <groupId>org.mskcc.cmo.metadb</groupId>
       <artifactId>persistence</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
ao508, line 23: can you move this log dependency to the master pom?

ao508: and the property as well

service/src/main/java/org/mskcc/cmo/metadb/service/MetaDbRequestService.java
```diff
@@ -7,7 +7,7 @@
  *
  * @author ochoaa
  */
-public interface CmoRequestService {
+public interface MetaDbRequestService {
```
ao508, line 10: 👍

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MetaDbRequestServiceImpl.java
```diff
@@ -59,6 +62,10 @@ public void saveRequest(MetaDbRequest request) throws Exception {
     @Override
     public Map<String, Object> getMetaDbRequest(String requestId) throws Exception {
         MetaDbRequest metaDbRequest = metaDbRequestRepository.findByRequestId(requestId);
+        if (metaDbRequest == null) {
+            LOG.info("Couldn't find a request with requestId " + requestId);
```
ao508, line 66: change to `LOG.error`

web/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+      <artifactId>springfox-swagger-ui</artifactId>
+      <version>2.7.0</version>
+      <scope>compile</scope>
+    </dependency>
+     <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web</artifactId>
+      <exclusions>
+        <exclusion>
+            <groupId>ch.qos.logback</groupId>
```
ao508, line 38: change the indentation for these 2 lines to 2 spaces, I think they're at 4 spaces right now

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,67 @@
...
+        method = RequestMethod.GET,
+        produces = "application/json")
+    @ResponseBody
+    public Map<String, Object> fetchMetaDbRequestGET(@ApiParam(value = 
+        "Retrieves MetaDbRequest from a RequestId.",
+        required = true)
+        @PathVariable String requestId) throws Exception {
+        return metaDbRequestService.getMetaDbRequest(requestId);
+    }
+
```
ao508, line 48: replace instances of "fetchRequestListPOST" with "fetchRequestPOST"

web/src/main/java/org/mskcc/cmo/metadb/web/RequestController.java
```diff
@@ -0,0 +1,67 @@
...
+        required = true)
+        @PathVariable String requestId) throws Exception {
+        return metaDbRequestService.getMetaDbRequest(requestId);
+    }
+    
+    /**
+     * fetchRequestListPOST
+     * @param requestIds
+     */
+    @ApiOperation(value = "Retrieves list of MetaDbRequest from a list of RequestId.",
```
ao508, line 53: "... list of request ids"

web/src/test/java/org/mskcc/cmo/metadb/web/RestIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
+package org.mskcc.cmo.metadb.web;
+
+import static org.junit.Assert.assertEquals;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
```
ao508, line 7: extra newlines here?

web/src/test/java/org/mskcc/cmo/metadb/web/RestIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
+package org.mskcc.cmo.metadb.web;
```
ao508, line 1: I forgot to mention that the standard naming convention for test classes is "<class name>Test" so we should rename this class "RequestControllerTest"

web/src/test/java/org/mskcc/cmo/metadb/web/RestIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private List<Map<String, Object>> fetchMetaDbRequestListPOST(String[] requestIds)
+    {
+        String responses = this.restTemplate.postForObject(BASE_URL, requestIds, String.class);
+        JsonParser springParser = JsonParserFactory.getJsonParser();
+        List<Map<String, Object>> lists = (List<Map<String, Object>>)(List<?>) springParser.parseList(responses);
+        return lists;
+    }
+    
+    @Test
+    public void testMetaDbRequestFetch() {
```
ao508, line 40: Is the test actually establishing a connection to the neo4j on dashidev?

If so then we should add an embedded neo4j graphdb (can you see if this is possible?) I'm sure there are others who want to test the same sort of functionality.. 

Anyway, if we can't create and populate a small neo4j graphdb in memory during testing then an alternative would be to use Mockito to mock responses [see genome nexus annotator example here](https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/master/annotator/src/test/java/org/cbioportal/annotator/MockGenomeNexusImpl.java)

ao508 (changes requested):

ao508:
Some questions I have & cases that the tests and controller should be able to handle:

- [ ] POST method should return `Map<String, Map<String, Object>>` (example here)[https://github.com/genome-nexus/genome-nexus/blob/master/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java#L48] where the key 
is the request id and the value is the `Map<String, Object>` representing the metadb request json
- [ ] The POST response should return the data for each given request id in the order that they were received. I believe the name of this type of class is something like "LinkedHashMap" - you may have to double check. 
- [ ] If a request ID in a POST response is invalid (it does not exist in the graphdb) then return a null but add the request id in the returned map

I'll let you know if I think of more ways the controller should behave.

web/src/test/java/org/mskcc/cmo/metadb/web/RestIntegrationTest.java
ao508: mock data can be loaded from the test data files that we checked into the repo: https://github.com/mskcc/cmo-metadb/tree/master/test/data

ao508: one basic test case: mock data loaded from one of the the json files linked above and check that it's deserialized correctly (fields match the expected values)

ao508 (approved):

PR #153, opened 2021-02-25
Missing attributes

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -43,6 +43,8 @@
     protected String libraryType;
     protected List<RequestSample> requestSamples;
     protected List<String> pooledNormals;
+    protected boolean isCmoRequest;
```
ao508, line 46: `isCmoRequest` is the field being removed and `cmoRequest` is what's staying I thought

ao508 (changes requested):

model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
divyamadala30: you're right, my bad

model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -90,6 +95,8 @@ public MetaDbRequest(String requestId, String recipe, String projectManagerName,
         this.metaDbSampleList = metaDbSampleList;
         this.metaDbProject = new MetaDbProject(requestId.split("_")[0]);
         this.requestJson = requestJson;
+        this.setBicAnalysis(bicAnalysis);
```
ao508, line 98: the constructor shouldnt be referencing the setter methods

ao508 (approved):

PR #176, opened 2021-03-10
Refactored creationTime to importDate

divyamadala30:
idSource is named to namespace

### DIFF
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
@@ -22,7 +22,7 @@ public class SampleManifestEntity implements Serializable {
     @JsonIgnore
     private Long id;
     private UUID sampleUuid;
-    private String creationTime;
+    private String importDate;
     private String cmoInfoIgoId;
     private String cmoSampleName;
     private String sampleName;
@@ -58,13 +58,12 @@ public class SampleManifestEntity implements Serializable {
     private String principalInvestigator;
     private String ancestorSample;
     private String sampleStatus;
-    private Date creationDate;
     private String requestId;
 
     public SampleManifestEntity() {}
 
-    public SampleManifestEntity(String creationTime) {
-        this.creationTime = creationTime;
+    public SampleManifestEntity(String importDate) {
+        this.importDate = importDate;
     }
 
     /**
@@ -101,7 +100,7 @@ public class SampleManifestEntity implements Serializable {
      * @param principalInvestigator
      * @param ancestorSample
      * @param sampleStatus
-     * @param creationTime
+     * @param importDate
      */
     public SampleManifestEntity(String igoId, String cmoInfoIgoId, String cmoSampleName, String sampleName,
             String cmoSampleClass, String cmoPatientId, String investigatorSampleId, String oncoTreeCode,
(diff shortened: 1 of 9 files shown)


ao508 (commented):
I checked out your branch locally and saw references to "idSource" still in 
```
persistence/src/main/java/org/mskcc/cmo/metadb/persistence/MetaDbRequestRepository.java
persistence/src/main/java/org/mskcc/cmo/metadb/persistence/MetaDbSampleRepository.java
```

model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
```diff
@@ -137,7 +137,7 @@ public SampleManifestEntity(String igoId, String cmoInfoIgoId, String cmoSampleN
         this.fastqPath = fastqPath;
         this.ancestorSample = ancestorSample;
         this.sampleStatus = sampleStatus;
-        this.creationTime = creationTime;
+        this.importDate = creationTime;
```
ao508, line 140: you should also update the parameter name to. match

model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
```diff
@@ -470,12 +471,12 @@ public void setSampleStatus(String sampleStatus) {
...
-    public Date getCreationDate() {
-        return creationDate;
+    public Date getLimsCreationDate() {
+        return limsCreationDate;
     }
 
-    public void setCreationDate(Date creationDate) {
-        this.creationDate = creationDate;
+    public void setLimsCreationDate(Date limsCreationDate) {
+        this.limsCreationDate = limsCreationDate;
```
ao508, line 479: I still dont see the lims creation date in the source json? where is this from?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -55,22 +55,22 @@ public MetaDbSample setUpMetaDbSample(MetaDbSample
...
     }
 
     @Override
     public MetaDbSample setUpSampleManifestEntity(MetaDbSample metaDbSample) throws Exception {
-        for (SampleManifestEntity s: metaDbSample.getSampleManifestList()) {
-            Timestamp time = Timestamp.from(Instant.now());
-            s.setCreationTime(String.valueOf(time.getTime()));
+        SampleManifestEntity s = metaDbSample.getSampleManifestList().get(0);
+        if (s != null) {
+            s.setImportDate(LocalDateTime.now().format(DateTimeFormatter.ISO_LOCAL_DATE));
```
ao508, line 73: note for self: ISO_LOCAL_DATE format is "YYYY-MM-DD"

ao508 (changes requested):

model/src/main/java/org/mskcc/cmo/metadb/model/SampleManifestEntity.java
divyamadala30: It came with lims data members for SampleManifest. The json published by lims doesn't have it so the value of limsCreationDate is set to null. Null data members are not added to the json.

ao508: okay let's remove this then

ao508: can you maybe link where in their code this is?

divyamadala30: I think we can remove ti since lims isn't sending this information in their json

ao508: oh that might have been something that _we_ introduced for our own graph db. Yes let's definitely remove  it then

ao508 (commented):
Code basically looks good to me now. Can you be sure to test this out and see that the graphdb looks like what's expected?

ao508 (approved):

PR #182, opened 2021-03-12
Init commit WORK IN PROGRESS - Make operations transactional

ao508:

### DIFF
--- a/persistence/pom.xml
+++ b/persistence/pom.xml
@@ -16,5 +16,15 @@
       <artifactId>model</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.data</groupId>
+      <artifactId>spring-data-jpa</artifactId>
+      <version>2.4.5</version>
+    </dependency>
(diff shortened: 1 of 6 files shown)


persistence/src/main/java/org/mskcc/cmo/metadb/persistence/Neo4jRepositoryImpl.java
```diff
@@ -0,0 +1,28 @@
+package org.mskcc.cmo.metadb.persistence;
```
ao508, line 1: currently not used in this version of the repository. but keeping it for reference in case the other method I am implementing the transactions fails

persistence/pom.xml
```diff
@@ -16,5 +16,15 @@
       <artifactId>model</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
ao508, line 19: we will not need these dependencies if I can implement the transactions the simple way through the `@Transactional` annotation.

PR #185, opened 2021-03-15
still work in progress

divyamadala30:


model/src/main/java/org/mskcc/cmo/metadb/model/Library.java
```diff
@@ -204,4 +206,49 @@ public Integer getNumFastQs() {
     public String toString() {
         return ToStringBuilder.reflectionToString(this);
     }
+    
+    /**
+     * 
+     * @param libraryList
+     * @return true is Object Library is found in the list of Libraries
+     */
+    public boolean existsIn(List<Library> libraryList) {
```
ao508, line 215: move towards introducing an overrides `equals(..)` method instead for these classes

remove list comparisons being done here

model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -166,6 +166,16 @@ public void addMetaDbSampleList(MetaDbSample metaDbSample) {
         }
         metaDbSampleList.add(metaDbSample);
     }
+
```
ao508, line 169: same comment here about the request. we should. just override the `equals()` and `hashCode` methods and _then_ worry about the list comparisons

ao508 (commented):
After quickly going through this draft we should def focus on just overriding the `equals()` and `hashCode()` methods discussed and then resolve the array list comparisons afterward

PR #198, opened 2021-03-18
Update save request

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/service/impl/MetaDbRequestServiceImpl.java
```diff
@@ -20,17 +25,25 @@
  */
 @Component
 public class MetaDbRequestServiceImpl implements MetaDbRequestService {
+    
+    @Value("${metadb.request_handling_failure}")
```
ao508, line 29: rename to `metadb.request_handling_failures_filepath`

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MetaDbRequestServiceImpl.java
```diff
@@ -45,17 +58,33 @@ public void saveRequest(MetaDbRequest request) throws Exception {
...
         } else {
-            for (MetaDbSample s: request.getMetaDbSampleList()) {
-                if (s.getSampleIgoId() != null
-                        && metaDbRequestRepository.findMetaDbSampleByRequestAndIgoId(request.getRequestId(),
-                        s.getSampleIgoId().getSampleId()) == null) {
-                    savedRequest.addMetaDbSampleList(s);
-                    metaDbRequestRepository.save(savedRequest);
-                }
-            }
+
```
ao508, line 63: I think in general it's best to not create and use something in the same line that has the potential to throw an exception

Separate the declaration of the File returned by `fileUtil.getOrCreateFileWithHeader()`  and then use the returned object in `fileUtil.writeToFile()`

ao508 (approved):

PR #202, opened 2021-03-19
Add namespace to project node, fix metadb id field names

ao508:


model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
```diff
@@ -18,7 +18,7 @@
     @Id @GeneratedValue
     @JsonIgnore
     private Long id;
-    private UUID sampleMetaDbId;
+    private UUID metaDbSampleId;
```
ao508, line 21: fixed the field names to be consistent with same fields in other models

n1zea144 (approved):

divyamadala30 (approved):

PR #231, opened 2021-04-19
Jetstream

divyamadala30:


persistence/src/main/java/org/mskcc/cmo/metadb/persistence/MetaDbPatientRepository.java
```diff
@@ -10,8 +10,8 @@
  * @author ochoaa
  */
 public interface MetaDbPatientRepository extends Neo4jRepository<MetaDbPatient, Long> {
-    @Query("MATCH (pm: Patient) "
-        + "WHERE $cmoPatientId = pm.cmoPatientId RETURN pm")
+    @Query("MATCH (pm: Patient)<-[:IS_ALIAS]-(pa:PatientAlias "
+            + "{value: $cmoPatientId}) RETURN pm")
```
n1zea144, line 14: Am I mistaken, or aside from the argument label `cmoPatientId`, this query looks like it can work with any alias.  Should we rename the method `findPatientByPatientAlias` ?

ao508: ah yeah that seems to be the case. Maybe change the input name to `$patientId`

divyamadala30:
Screen Shot 2021-04-21 at 9 07 07 AM

ao508 (approved):

PR #303, opened 2021-06-16
Add endpoint to return samples given a cmo patient id

ao508:
This pr also includes changes to:
- handle a new field being sent as part of the SampleManifest JSON called "cmoSampleIdFields"
- update the metadb commons commit hash
- update the metadb messaging library commit hash

**Relevant screen shots**

format of "cmoSampleIdFields"

Screen Shot 2021-06-16 at 1 47 56 PM

**how cmo sample id fields looks like in neo4j**

Screen Shot 2021-06-16 at 1 48 36 PM

**how the new endpoint looks on the swagger page**

Screen Shot 2021-06-16 at 1 50 50 PM

### DIFF
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
@@ -5,9 +5,11 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.mskcc.cmo.metadb.model.converter.LibrariesStringConverter;
+import org.mskcc.cmo.metadb.model.converter.MapStringConverter;
 import org.mskcc.cmo.metadb.model.converter.QcReportsStringConverter;
 import org.neo4j.ogm.annotation.GeneratedValue;
 import org.neo4j.ogm.annotation.Id;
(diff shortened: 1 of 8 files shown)


model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
```diff
@@ -101,7 +102,7 @@ public SampleMetadata(String igoId, String cmoInfoIgoId, String cmoSampleName, S
             String tumorOrNormal, String tissueLocation, String specimenType, String sampleOrigin,
             String preservation, String collectionYear, String sex, String species, String tubeId,
             String cfDNA2dBarcode, List<QcReport> qcReports, List<Library> libraries,
-            String mrn, String cmoSampleId, String sampleType, String tumorType, String parentTumorType,
```
ao508, line 104: `cmoSampleId` wasn't being set or used anywhere so I removed it. I think this might have been added by accident to be used as `cmoSampleName` (the generated CMO sample label)

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/MetaDbSampleRepository.java
```diff
@@ -61,8 +61,16 @@
 
     @Query("MATCH (r: Request {requestId: $reqId}) "
             + "MATCH(r)-[:HAS_SAMPLE]->(sm: Sample) "
-            + "MATCH (sm)<-[:IS_ALIAS]-(s: SampleAlias {toLower(namespace): 'igoid', value: $igoId}) "
+            + "MATCH (sm)<-[:IS_ALIAS]-(s: SampleAlias {namespace: 'igoId', value: $igoId}) "
```
ao508, line 64: this `toLower` neo4j method was throwing an exception so I've removed it. Since we're setting this field I think it's safe to just refer to the namespace in its proper casing

divyamadala30 (approved):

PR #316, opened 2021-06-24
Remove sanity checking

ao508:
⚠️ **TODO: Need to decide on whether we still want to have logging done in cmo metadb to record duplicate requests that come through** 
----

https://github.com/mskcc/cmo-metadb/pull/316/files#diff-2fee701fd9cface3562b4c12c46a0c818112e6f5d522fd30719778d703b54369R69

Alternative solution might be to publish to a "logging" topic that will be handled by the request filter/sample validator which will then log the request.

### DIFF
--- a/service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
+++ b/service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
@@ -26,8 +26,6 @@ import org.mskcc.cmo.metadb.model.MetaDbSample;
 import org.mskcc.cmo.metadb.model.SampleMetadata;
 import org.mskcc.cmo.metadb.service.MessageHandlingService;
 import org.mskcc.cmo.metadb.service.MetaDbRequestService;
-import org.mskcc.cmo.metadb.service.util.RequestStatusLogger;
-import org.neo4j.driver.internal.shaded.io.netty.util.internal.StringUtil;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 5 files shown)


ao508:
@divyamadala30 the review can wait until the request-reply messaging is implemented in the messaging lib

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -178,12 +158,6 @@ public void onMessage(Message msg, Object message) {
                     messageHandlingService.newRequestHandler(metaDbRequest);
                 } catch (Exception e) {
                     LOG.error("Exception during processing of request on topic: " + IGO_NEW_REQUEST_TOPIC, e);
-                    try {
-                        requestStatusLogger.logRequestStatus(message.toString(),
-                                RequestStatusLogger.StatusType.REQUEST_PARSING_ERROR);
```
divyamadala30, line 183: Shouldn't we log parsing errors

divyamadala30 (commented):
Question about using timestamps to find duplicates in RequestServiceImpl: Are we still using that even after setting up ids to catch duplicates?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
ao508: those would be handled in the request filter app - I have a separate PR w/some changes to that

ao508: @divyamadala30 here it is https://github.com/mskcc/cmo-metadb-request-filter/pull/6/files

divyamadala30: Doesn't this catch errors while converting the message into a java object?

divyamadala30 (approved):

PR #323, opened 2021-07-08
metadb can now handle request/reply

divyamadala30:

### DIFF
--- a/service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
+++ b/service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
@@ -1,0 +1,167 @@
+package org.mskcc.cmo.metadb.service.impl;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.nats.client.Message;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.Phaser;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.mskcc.cmo.messaging.Gateway;
+import org.mskcc.cmo.messaging.MessageConsumer;
+import org.mskcc.cmo.metadb.model.SampleMetadata;
+import org.mskcc.cmo.metadb.service.RequestReplyHandlingService;
+import org.mskcc.cmo.metadb.service.SampleService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RequestReplyHandlingServiceImpl implements RequestReplyHandlingService {
+    
+    @Value("${request_reply.patient_samples_topic}")
+    private String REQUEST_REPLY_TOPIC;
+
+    @Value("${num.new_request_handler_threads}")
+    private int NUM_NEW_REQUEST_HANDLERS;
+
+    @Autowired
+    private SampleService sampleService;
+
+    private final ObjectMapper mapper = new ObjectMapper();
+    private static boolean initialized = false;
+    private static volatile boolean shutdownInitiated;
+    private static final ExecutorService exec = Executors.newCachedThreadPool();
+    private static final BlockingQueue<ReplyInfo> newRequestReplyQueue =
+        new LinkedBlockingQueue<ReplyInfo>();
+    private static CountDownLatch newRequestHandlerShutdownLatch;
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        final Phaser newSamplePhaser = new Phaser();
+        newSamplePhaser.register();
+        for (int lc = 0; lc < NUM_NEW_REQUEST_HANDLERS; lc++) {
+            newSamplePhaser.register();
+            exec.execute(new NewRequestReplyHandler(newSamplePhaser));
+        }
+        newSamplePhaser.arriveAndAwaitAdvance();
+    }
+    
+}
```
divyamadala30, line 162: Need to add an extra line here

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.nats.client.Message;
+
+@Component
+public class RequestReplyHandlingServiceImpl implements RequestReplyHandlingService {
+    
+    @Value("${request_reply.patient_samples_topic}")
```
ao508, line 27: don't forget to add this new prop to the example properties

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        @Override
+        public void run() {
+            phaser.arrive();
+            while (true) {
+                try {
+                    ReplyInfo replyInfo = newRequestReplyQueue.poll(100, TimeUnit.MILLISECONDS);
+                    if ((replyInfo != null)) {
+                        System.out.println("Publishing to: " + replyInfo.getReplyTo() + "\n message: "+ replyInfo.getReplyMessage());
```
n1zea144, line 82: Lets get rid of this output.  If we want it, we should put into a log file.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+import org.mskcc.cmo.metadb.service.RequestReplyHandlingService;
+import org.mskcc.cmo.metadb.service.SampleService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.nats.client.Message;
+
+@Component
+public class RequestReplyHandlingServiceImpl implements RequestReplyHandlingService {
```
n1zea144, line 25: Is the idea that this class will contain all request-reply handlers and MessagingHandlingService will contain all other handlers?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+    
+    @Value("${request_reply.patient_samples_topic}")
+    private String REQUEST_REPLY_TOPIC;
```
n1zea144, line 28: Suggest this be called something more specific to the actual topic in the event a new request-reply topic is introduced in the future.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
n1zea144, line 82: Lets get rid of the output to stdout. If we want it, we should put into a log file.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestReplyHandlingServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        }        
+    }
+
+    @Override
+    public void newPatientSamplesHandler(String patientId, String replyTo) throws Exception {
+        if (!initialized) {
+            throw new IllegalStateException("Message Handling Service has not been initialized");
+        }
+        if (!shutdownInitiated) {
+            List<SampleMetadata> sampleList = sampleService.getSampleMetadataListByCmoPatientId(patientId);
```
n1zea144, line 117: I would move this into NewRequestReplyHandler.run() so we don't tie up the message handling thread.

n1zea144 (changes requested):

n1zea144 (approved):

PR #364, opened 2021-08-27
RequestService can now handle updates to request metadata 

divyamadala30:
does not persist to NEO4J

### DIFF
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
@@ -37,7 +37,7 @@ public class MetaDbRequest implements Serializable {
     private MetaDbProject metaDbProject;
     @JsonIgnore
     @Relationship(type = "HAS_METADATA", direction = Relationship.OUTGOING)
-    private RequestMetadata requestMetadata;
+    private List<RequestMetadata> requestMetadataList;
     @JsonIgnore
     private String namespace;
     // need this field to deserialize message from IGO_NEW_REQUEST properly
@@ -139,12 +139,23 @@ public class MetaDbRequest implements Serializable {
         this.metaDbProject = metaDbProject;
     }
     
-    public RequestMetadata getRequestMetadata() {
-        return requestMetadata;
+    public List<RequestMetadata> getRequestMetadataList() {
+        return requestMetadataList;
     }
 
-    public void setRequestMetadata(RequestMetadata requestMetadata) {
-        this.requestMetadata = requestMetadata;
+    public void setRequestMetadataList(List<RequestMetadata> requestMetadataList) {
+        this.requestMetadataList = requestMetadataList;
+    }
+    
+    /**
+     * 
+     * @param requestMetadata
+     */
+    public void addRequestMetadata(RequestMetadata requestMetadata) {
+        if (requestMetadataList == null) {
+            requestMetadataList = new ArrayList<>();
+        }
+        requestMetadataList.add(requestMetadata);
     }
 
     public String getNamespace() {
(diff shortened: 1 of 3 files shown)


model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -311,6 +337,32 @@ public boolean getBicAnalysis() {
...
+        this.labHeadEmail = requestMetadataMap.get("labHeadEmail");
+        this.investigatorName = requestMetadataMap.get("investigatorName");
+        this.investigatorEmail = requestMetadataMap.get("investigatorEmail");
+        this.dataAnalystName = requestMetadataMap.get("dataAnalystName");
+        this.otherContactEmails = requestMetadataMap.get("otherContactEmails");
+        this.dataAccessEmails = requestMetadataMap.get("dataAccessEmails");
+        this.qcAccessEmails = requestMetadataMap.get("qcAccessEmails");
+        this.strand = requestMetadataMap.get("strand");
+        this.libraryType = requestMetadataMap.get("libraryType");
+        this.metaDbProject = new MetaDbProject(requestMetadataMap.get("requestId").split("_")[0]);
```
ao508, line 362: the project the request belongs to wouldnt change so this line could be omitted

model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -131,6 +138,25 @@ public MetaDbProject getMetaDbProject() {
...
+    public List<RequestMetadata> getRequestMetadataList() {
+        return requestMetadataList;
+    }
+
+    public void setRequestMetadataList(List<RequestMetadata> requestMetadataList) {
+        this.requestMetadataList = requestMetadataList;
+    }
+    
+    /**
+     *
```
ao508, line 151: if adding javadoc comments then a brief description of what the method does should be added as well

model/src/main/java/org/mskcc/cmo/metadb/model/MetaDbRequest.java
```diff
@@ -311,6 +337,32 @@ public boolean getBicAnalysis() {
     public void setBicAnalysis(boolean bicAnalysis) {
         this.bicAnalysis = bicAnalysis;
     }
+    
+    /**
+     *
```
ao508, line 342: same comment here re: javadoc comments

model/src/main/java/org/mskcc/cmo/metadb/model/RequestMetadata.java
```diff
@@ -0,0 +1,43 @@
...
+import org.neo4j.ogm.annotation.Id;
+
+public class RequestMetadata {
+    @Id @GeneratedValue
+    private Long id;
+    private String requestMetadataJson;
+    private String importDate;
+    
+    /**
+     *
```
ao508, line 14: javadoc comment

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MetaDbRequestServiceImpl.java
```diff
@@ -120,4 +129,42 @@ private Boolean timestampWithin24Hours(Date referenceTimestamp, Date newTimestam
...
+        Map<String, String> updatedRequestMap = mapper.readValue(requestMetadataJson.toString(), Map.class);
+         
+        for (Map.Entry<String,String> entry : updatedRequestMap.entrySet()) {
+            currentRequestMap.replace(entry.getKey(), entry.getValue());
+        }
+        
+        return mapper.writeValueAsString(currentRequestJson);
+    }
+    
+    private void updateRequestMetadata(String requestMetadataJson)
```
ao508, line 156: if I understand correctly then this method would be called if we already have an existing request in the database? So it would probably be called [in this else block](https://github.com/mskcc/cmo-metadb/pull/364/files#diff-2fee701fd9cface3562b4c12c46a0c818112e6f5d522fd30719778d703b54369R77)

Maybe the `MetadbRequest.addRequestMetadata()` should not just append the list but also call its other method `updateRequestMetadata()` internally so that in one step you are 
1. updating the requestMetadataList
2. updating the `MetadbRequest` data members
3. maybe also update the value of `MetadbRequest.requestJson` (the json string?)

I think that would allow us to get rid of this method here. It Also would allow us to change the inputs of `metadbRequest.updateRequestMetadata()` since `metadbRequest` object would already have access to its current request json string

Thoughts?

If we move forward with this then I wouldn't worry about implementing these changes for this PR specifically. We can have it in a separate PR.

ao508 (commented):
I made a few comments if you could take a moment to go through and answer some questions/thoughts that I had. I think this PR is ready for merge though so once those questions are answered then I'll approve/merge. These notes will be helpful with refining the work done here already.

ao508 (approved):

PR #368, opened 2021-09-01
Support request data versioning

ao508:
- further development for supporting request data versioning
- fixed some issues with metadb sample java obj


service/src/main/java/org/mskcc/cmo/metadb/service/impl/MetaDbRequestServiceImpl.java
```diff
@@ -73,31 +73,68 @@ public boolean saveRequest(MetaDbRequest request) throws Exception {
                 request.setMetaDbSampleList(updatedSamples);
             }
             requestRepository.save(request);
-            return true;
+            return Boolean.TRUE;
         } else {
```
divyamadala30, line 77: How does it handle updatedRequestMetadata where the nats message is just the metadata json?

PR #375, opened 2021-09-13
Mockdb

ao508:
TODO:
- Still need to add further unit tests but the data loads and is persisted to the embedded database now. All tests for the service layer should live in `MetadbServiceTest.java`


persistence/src/test/java/org/mskcc/cmo/metadb/persistence/MockNeo4jDbTest.java
```diff
@@ -0,0 +1,101 @@
...
+                MetaDbRequest.class);
+        metaDbRequest.setRequestJson(requestJson);
+        metaDbRequest.setMetaDbSampleList(extractMetaDbSamplesFromIgoResponse(requestJson));
+        metaDbRequest.setNamespace("igo");
+
+        requestRepository.save(metaDbRequest);
+
+        List<MetaDbSample> sampleList = sampleRepository
+                .findAllMetaDbSamplesByRequest(metaDbRequest.getRequestId());
+        Assertions.assertTrue(sampleList.size() == 4);
```
ao508, line 81: - use one of these MetaDbSample objects, make a copy of its SampleMetadata and make changes to the SampleMetadata. 
- Persist the changed/updated `SampleMetadata` to the database and then query for the same sample's `SampleMetadata` history to make sure that it now includes 2 versions of the data

PR #379, opened 2021-09-16
Testcontainers CircleCI fix

ao508:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,9 @@
-defaults: &defaults
-    docker:
-        - image: cimg/openjdk:8.0.275
-
 version: 2.1
-
-orbs:
-  maven: circleci/maven@1.0.3
 
 jobs:
   run_checkstyle:
-    <<: *defaults
     working_directory: /tmp/repos/cmo-metadb
+    machine: true
     steps:
       - checkout
       - run:
@@ -24,11 +17,8 @@ jobs:
               git fetch --tags && \
               git checkout ${METADB_COMMON_VERSION}
       - run:
-          name: "Build project..."
-          command: mvn clean install
-      - run:
-          name: "Run checkstyle plugin..."
-          command: mvn checkstyle:checkstyle
+          name: "Build project and run checkstyle plugin..."
+          command: mvn clean install checkstyle:checkstyle
       - run:
           name: "Validating checkstyle plugin report..."
           command: bash /tmp/repos/cmo-metadb-common/test/scripts/generate_checkstyle_report.sh /tmp/repos/cmo-metadb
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -7,11 +7,22 @@ version: 2.1
...
+    environment:
+      - TESTCONTAINERS_RYUK_DISABLED: true
     steps:
+      - setup_remote_docker:
+          version: 20.10.2
+      - run:
+          name: "Configure testcontainers.properties..."
+          command: |
+            touch /home/circleci/.testcontainers.properties && \
+            echo "DOCKER_HOST = unix:///var/run/docker.sock" > /home/circleci/.testcontainers.properties
```
ao508, line 25: just trying things out here. `testcontainers.properties`, if we actually need it, can be checked into `src/test/resources`

.circleci/config.yml
```diff
@@ -7,11 +7,22 @@ version: 2.1
...
+
 jobs:
   run_checkstyle:
     <<: *defaults
     working_directory: /tmp/repos/cmo-metadb
+    environment:
+      - TESTCONTAINERS_RYUK_DISABLED: true
     steps:
+      - setup_remote_docker:
+          version: 20.10.2
```
bsideup, line 20: you don't need a remote Docker if you are using `machine: true`. In fact, that's most probably why it doesn't work :)
```suggestion
```

.circleci/config.yml
bsideup, line 25: not needed. Testcontainers will use `unix:///var/run/docker.sock` automagically.
```suggestion
```

kiview: This should normally not be necessary and Testcontainers should detect the Docker environment by itself.

bsideup:
Hey @ao508, I just left a couple of comments. I think the problem was with a mix of `machine: true` and an unnecessary `setup_remote_docker`. See Testcontainers' docs on how to use Testcontainers with CircleCI: https://www.testcontainers.org/supported_docker_environment/continuous_integration/circle_ci/

.circleci/config.yml
```diff
@@ -7,11 +7,22 @@ version: 2.1
...
   maven: circleci/maven@1.0.3
 
+machine: true
+
 jobs:
   run_checkstyle:
     <<: *defaults
     working_directory: /tmp/repos/cmo-metadb
+    environment:
+      - TESTCONTAINERS_RYUK_DISABLED: true
```
bsideup, line 17: also unnecessary:
```suggestion
```

.circleci/config.yml
ao508: thanks! yeah I've tried several variations with and without machine + with/without installing docker manually

ao508: i *do* think that the solution is getting the right docker installation going inside of the container, will update when I've played around with it a bit more

ao508: and D O N E 🙌

PR #380, opened 2021-09-16
Unit tests for sampleService

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/service/SampleService.java
```diff
@@ -19,7 +19,8 @@
 
     MetaDbSample getMetaDbSample(UUID metaDbSampleId) throws Exception;
 
-    MetaDbSample getMetaDbSampleByRequestAndIgoId(String requestId, SampleAlias igoId) throws Exception;
+    MetaDbSample getMetaDbSampleByRequestAndIgoSampleAlias(String requestId,
```
n1zea144, line 22: Can we drop Igo from the method name?

n1zea144 (approved):
We are probably going to want to search for a sample/history by its alias, but I don't think we need to build it in until that comes up.

service/src/main/java/org/mskcc/cmo/metadb/service/SampleService.java
ao508: maybe `getMetaDbSampleByRequestAndIgoSampleAlias` --> `getMetaDbSampleByRequestAndAlias` ?

ao508 (changes requested):
Can you add java docs to each of the Test methods explaining what they're testing for?

service/src/test/java/org/mskcc/cmo/metadb/service/MockDataUtils.java
```diff
@@ -88,7 +88,7 @@ public MetaDbRequest extractRequestFromJsonData(String requestJson) throws Excep
     }
 
     /**
-     * Extracts a List of MetaDbSample's given a request json string;
+     * Extracts a List<MetaDbSample> given a request json string
```
ao508, line 91: this is where the checkstyle issue is failing, i would just revert it to what it was previously

ao508 (approved):

PR #392, opened 2021-09-22
Bug fix for duplicate patient nodes

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -42,6 +42,7 @@ public MetaDbSample saveSampleMetadata(MetaDbSample
             if (patient != null) {
                 updatedMetaDbSample.setPatient(patient);
             }
+            patientRepository.save(updatedMetaDbSample.getPatient());
```
ao508, line 45: does this not return a `MetaDbPatient` object which would have a UUID assigned to it?

If not then that's probably what we want. Also the save should only happen if the patient was not found in the db so it should be placed in an `else` block for this `if` statement

service/src/test/java/org/mskcc/cmo/metadb/service/MetadbServiceTest.java
```diff
@@ -1,10 +1,16 @@
 package org.mskcc.cmo.metadb.service;
 
+
+import static org.junit.Assert.assertThrows;
```
ao508, line 4: this is going to fail the checkstyle validation

service/src/test/java/org/mskcc/cmo/metadb/service/MetadbServiceTest.java
```diff
@@ -207,5 +213,26 @@ public void testSampleHistoryListIsAscendingByImportDate() throws Exception {
...
+    public void testFindPatientByPatientAliasWithExpectedFailure() {
+        MetaDbPatient patient = new MetaDbPatient();
+        PatientAlias patientAlias = new PatientAlias();
+        patientAlias.setPatientId("C-MP789JR");
+        patient.addPatientAlias(patientAlias);
+        patientRepository.save(patient);
+        
+        try {
+            MetaDbPatient retrievedPatient = patientRepository.findPatientByPatientAlias("C-MP789JR");
+            Assert.fail();
```
ao508, line 232: you can remove this try/catch black and instead add an annotation to this test method that indicates it's supposed to throw a specific exception (not just a generic Exception)

the annotation is something like `@ExpectedException` or something like that

ao508 (changes requested):

service/src/test/resources/data/incoming_requests/mocked_request2a_one_normal.json
```diff
@@ -27,7 +27,7 @@
       "cmoSampleName": "C-TX6DNG-N001-d",
       "sampleName": "LMNO_4396_N",
       "cmoSampleClass": "Normal",
-      "cmoPatientId": "C-TX6DNG",
+      "cmoPatientId": "C-MP789JR",
```
ao508, line 30: I wouldnt change the value here, I would make a copy of this json data file and then change the patient id in the new file. We still want this mock dataset to be the same as it is currently

ao508 (changes requested):

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -41,6 +41,8 @@ public MetaDbSample saveSampleMetadata(MetaDbSample
                     updatedMetaDbSample.getPatient().getCmoPatientId().getPatientId());
             if (patient != null) {
                 updatedMetaDbSample.setPatient(patient);
+            } else {
+                patientRepository.save(updatedMetaDbSample.getPatient());
```
ao508, line 45: Still need to save this to a new object

service/src/test/resources/data/incoming_requests/mocked_request2c_shared_patient.json
```diff
@@ -0,0 +1,82 @@
...
+  "qcAccessEmails": "",
+  "pooledNormals": [
+    "/FASTQ/Project_POOLEDNORMALS/Sample_FFPEPOOLEDNORMAL_IGO_GENESET101_TGCTAGCT/FFPEPOOLEDNORMAL_IGO_GENESET101_TGCTAGCT_S118_R1_001.fastq.gz",
+    "/FASTQ/Project_POOLEDNORMALS/Sample_FFPEPOOLEDNORMAL_IGO_GENESET101_TGCTAGCT/FFPEPOOLEDNORMAL_IGO_GENESET101_TGCTAGCT_S118_R2_001.fastq.gz",
+    "/FASTQ/Project_POOLEDNORMALS/Sample_FROZENPOOLEDNORMAL_IGO_GENESET101_GCGGCTAG/FROZENPOOLEDNORMAL_IGO_GENESET101_GCGGCTAG_S22_R1_001.fastq.gz",
+    "/FASTQ/Project_POOLEDNORMALS/Sample_FROZENPOOLEDNORMAL_IGO_GENESET101_GCGGCTAG/FROZENPOOLEDNORMAL_IGO_GENESET101_GCGGCTAG_S22_R2_001.fastq.gz"
+  ],
+  "projectId": "22022",
+  "samples": [
+    {
```
ao508, line 25: there appears to only be one sample but the case that we want to test for is at least 2 samples with the same patient id and same new request

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: need to return foundSample in the else block below (in the case that the foundSample exists)

service/src/test/resources/data/incoming_requests/mocked_request2c_shared_patient.json
ao508: changes:
- different request id and project id that we do not already have in our mock datasets (make sure they start w/same pattern)
- add 2 more samples to this `samples` list, total should be 3 sample manifest jsons in this list
- all 3 samples share the same cmo patient id
- all 3 samples have different igo ids and cmo sample names

service/src/test/java/org/mskcc/cmo/metadb/service/MetadbServiceTest.java
```diff
@@ -207,5 +210,24 @@ public void testSampleHistoryListIsAscendingByImportDate() throws Exception {
...
     }
+    
+    @Test
+    public void testFindPatientByPatientAlias() {
+        Assertions.assertThat(patientRepository.findPatientByPatientAlias("C-MP789JR")).isNotNull();
+    }
+    
+    @Test
+    public void testFindPatientByPatientAliasWithExpectedFailure() {
+        MetaDbPatient patient = new MetaDbPatient();
```
ao508, line 221: related but not a change that is needed here:

- add a patient service that returns an updated metadb patient obj

ao508: sample service would use the new patientService instead of the patientRepository directly

ao508 (changes requested):

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: after implementing and replacing the use of the `patientRepository` here with the `patientService` have the service return the updated metadb patient obj and set THAT as the `updatedMetadbSample.setPatient(updatedPatient)`

PR #468, opened 2021-11-10
Update returned json

divyamadala30:


service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -107,12 +111,14 @@ public MetadbSample getMetadbSample(UUID metadbSampleId) throws Exception {
...
-                igoId.getSampleId());
+                igoId.getValue());
+        if (sample == null) {
+            return null;
+        }
         sample.setSampleMetadataList(sampleRepository
                 .findSampleMetadataListBySampleId(sample.getMetaDbSampleId()));
         for (SampleMetadata s : sample.getSampleMetadataList()) {
             s.setMetaDbPatientId(patientService.getPatientIdBySample(sample.getMetaDbSampleId()));
-            s.setMetaDbSampleId(sample.getMetaDbSampleId());
```
ao508, line 115: maybe instead of metadb patient id being set just above this we should just be fetching the actual `MetadbPatient` and calling `sample.setPatient(...)`

I feel it's a little redundant to keep access to metadb patient id on the sample object level while we also have access to the `MetadbPatient` itself at the sample object level as well

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -0,0 +1,503 @@
...
+            String cmoSampleClass, String cmoPatientId, String investigatorSampleId, String oncoTreeCode,
+            String tumorOrNormal, String tissueLocation, String specimenType, String sampleOrigin,
+            String preservation, String collectionYear, String sex, String species, String tubeId,
+            String cfDNA2dBarcode, List<QcReport> qcReports, List<Library> libraries,
+            String mrn, String sampleType, String tumorType, String parentTumorType,
+            String tissueSource, String recipe, String baitSet, String fastqPath,
+            String principalInvestigator, String ancestorSample, String sampleStatus, String importDate,
+            List<SampleAlias> sampleAliases) {
+        this.mrn = mrn;
+        this.setCmoInfoIgoId(cmoInfoIgoId);
```
ao508, line 154: this doesnt seem right

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -0,0 +1,503 @@
...
+        this.sampleOrigin = sampleOrigin;
+        this.tissueSource = tissueSource;
+        this.tissueLocation = tissueLocation;
+        this.recipe = recipe;
+        this.baitSet = baitSet;
+        this.principalInvestigator = principalInvestigator;
+        this.fastqPath = fastqPath;
+        this.ancestorSample = ancestorSample;
+        this.sampleStatus = sampleStatus;
+        this.setImportDate(importDate);
```
ao508, line 178: ?

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
ao508: *referring to calling the `set` method here

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -168,4 +176,10 @@ public Boolean sampleHasMetadataUpdates(SampleMetadata existingSampleMetadata,
         }
         return Boolean.FALSE;
     }
+
+    @Override
+    public PublishedMetadbSample getPublishedMetadbSample(UUID metadbSampleId) throws ParseException {
+        MetadbSample sample = sampleRepository.findSampleById(metadbSampleId);
```
ao508, line 182: remember that the sample object returned here is not going to have all of its relationships actually filled in which is why we have to call a method specifically when calling `saveSampleMetadata(...)` in the sample service to fill in the missing details. 

So for example this `MetadbSample` returned by this method isn't going to have any of its sample aliases linked and the `metadbPatientId` and `metadbSampelId` values aren't going to be populated either.

ao508 (changes requested):
These changes need to have some sanity checking done through use of the endpoint. Does `PublishedMetadbSample` look the way we expect it to w/the sample aliases, metadb patient id, and metadb sample id filled in?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -94,10 +96,12 @@ public MetadbSample fetchAndLoadSampleDetails(MetadbSample sample) throws Except
     @Override
     public MetadbSample getMetadbSample(UUID metadbSampleId) throws Exception {
         MetadbSample sample = sampleRepository.findSampleById(metadbSampleId);
+        if (sample == null) {
+            return null;
+        }
         sample.setSampleMetadataList(sampleRepository.findSampleMetadataListBySampleId(metadbSampleId));
         for (SampleMetadata s: sample.getSampleMetadataList()) {
             s.setMetaDbPatientId(patientService.getPatientIdBySample(metadbSampleId));
```
ao508, line 104: so after doing some of my own testing as well i confirmed that you don't need the patient uuid at the sample **metadata** level so we can remove this for-block (and others like this throughout the sample service) and instead just fetching and setting the patient of the `MetadbSample`

```
MetadbPatient patient = patientService.getPatientByCmoPatientId(cmoPatientId);
sample.setPatient(patient);
```

ao508 (changes requested):

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestServiceImpl.java
```diff
@@ -159,9 +160,9 @@ public PublishedMetadbRequest getPublishedMetadbRequestById(String requestId) th
         MetadbRequest request = getMetadbRequestById(requestId);
 
         // for each metadb sample get the latest version of its sample metadata
-        List<SampleMetadata> samples = new ArrayList<>();
+        List<PublishedMetadbSample> samples = new ArrayList<>();
         for (MetadbSample sample : request.getMetaDbSampleList()) {
-            samples.add(sample.getLatestSampleMetadata());
+            samples.add(sampleService.getPublishedMetadbSample(sample.getMetaDbSampleId()));
```
ao508, line 165: let's move the call to the service outside of the list modification

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -1,6 +1,7 @@
 package org.mskcc.cmo.metadb.service.impl;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.text.ParseException;
```
ao508, line 4: be sure to clean up any unused imports

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -171,4 +166,15 @@ public Boolean sampleHasMetadataUpdates(SampleMetadata existingSampleMetadata,
         }
         return Boolean.FALSE;
     }
+
+    @Override
+    public PublishedMetadbSample getPublishedMetadbSample(UUID metadbSampleId) throws Exception {
+        MetadbSample sample = getMetadbSample(metadbSampleId);
+        PublishedMetadbSample publishedSample = new PublishedMetadbSample(sample);
```
ao508, line 173: here we are passing the `MetadbSample` along to the `PublishedMetadbSample` constructor and then making a subsequent call to `publishedSample.setMetaDbSampleId()` 

since `MetadbSample` is being passed to the constructor anyway you can set the metadb sample id within the constructor itself instead of making this extra call here

ao508: same comment for setting up the patient stuff - all of this information is already available to the constructor

ao508: When calling [getMetadbSample](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-14fbd1dc6e948619c6d34efb8f41005965101954a8895440857979d916eb6b57R172) does it not also get and set the `MetadbPatient`?

I see metadb patient id [isn't getting set anymore](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-14fbd1dc6e948619c6d34efb8f41005965101954a8895440857979d916eb6b57L99) which i agree with but i did make a comment somewhere suggesting that we still set up the `MetadbPatient` field for the `MetadbSample` anywhere that we are constructing an instance of `MetadbSample`

ao508: once that is implemented then the patient service wouldnt need to be called in this method and we can move forward with the constructor containing all of the initialization/set up logic for creating a `PublishedMetadbSample`

ao508 (changes requested):

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -94,11 +96,10 @@ public MetadbSample fetchAndLoadSampleDetails(MetadbSample sample) throws Except
...
     public MetadbSample getMetadbSample(UUID metadbSampleId) throws Exception {
         MetadbSample sample = sampleRepository.findSampleById(metadbSampleId);
-        sample.setSampleMetadataList(sampleRepository.findSampleMetadataListBySampleId(metadbSampleId));
-        for (SampleMetadata s: sample.getSampleMetadataList()) {
-            s.setMetaDbPatientId(patientService.getPatientIdBySample(metadbSampleId));
-            s.setMetaDbSampleId(metadbSampleId);
+        if (sample == null) {
+            return null;
         }
+        sample.setSampleMetadataList(sampleRepository.findSampleMetadataListBySampleId(metadbSampleId));
```
ao508, line 102: need to fetch the sample's patient here still

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: still need to fetch and set the sample's patient here 

[reference](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-14fbd1dc6e948619c6d34efb8f41005965101954a8895440857979d916eb6b57L130)

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: [reference](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-14fbd1dc6e948619c6d34efb8f41005965101954a8895440857979d916eb6b57L130)

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: [reference](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-14fbd1dc6e948619c6d34efb8f41005965101954a8895440857979d916eb6b57L130)

ao508 (changes requested):

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -0,0 +1,506 @@
...
+
+    public void setCmoSampleIdFields(Map<String, String> cmoSampleIdFields) {
+        this.cmoSampleIdFields = cmoSampleIdFields;
+    }
+
+    public void setSampleAliases(List<SampleAlias> sampleAliases) {
+        this.sampleAliases = sampleAliases;
+    }
+
+    public List<SampleAlias> getSampleAliases() {
```
ao508, line 498: `getSampelAliases` should have same logic as other getters that return lists like [here](https://github.com/mskcc/cmo-metadb/pull/468/files#diff-34b0ee699f8dfd66575470b982c99631815ba2d9bd342263d973e14e82e0e8f6R277)

ao508 (changes requested):
A few more changes requested to `PublishedMetadbSample` - refer to comments 

Things to double check:

- [x] output json looks the way it's expected when fetching a request json by id (check swagger model and response when querying by id)
- [ ] add screenshot to this pr with a snippet of the response model featuring the added sample aliases in the sample metadata
- [ ] update sample controller to return List<PublishedMetadbSample> - this is something we missed/didnt consider

divyamadala30:
Example JSON format of publishedMetadbSample:
[example-published-request.txt](https://github.com/mskcc/cmo-metadb/files/7521660/example-published-request.txt)
Screen Shot 2021-11-12 at 10 54 02 AM

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -0,0 +1,513 @@
...
+    private String sampleStatus;
+    private String requestId;
+    @Convert(MapStringConverter.class)
+    private Map<String, String> cmoSampleIdFields;
+    private List<SampleAlias> sampleAliases;
+
+    public PublishedMetadbSample() {}
+
+    /**
+     * PublishedMetadbRequest constructor
```
ao508, line 66: `PublishedMetadbRequest` --> `PublishedMetadbSample`

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbSampleService.java
```diff
@@ -21,4 +22,6 @@ MetadbSample getMetadbSampleByRequestAndAlias(String requestId,
     List<SampleMetadata> getSampleMetadataHistoryByIgoId(String igoId) throws Exception;
     Boolean sampleHasMetadataUpdates(SampleMetadata existingSampleMetadata, SampleMetadata sampleMetadata)
             throws Exception;
+    PublishedMetadbSample getPublishedMetadbSamplebyUUID(UUID metadbSampleId) throws Exception;
```
ao508, line 25: `getPublishedMetadbSamplebyUUID` --> `getPublishedMetadbSample`

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -168,4 +170,26 @@ public Boolean sampleHasMetadataUpdates(SampleMetadata existingSampleMetadata,
...
     }
+
+    @Override
+    public PublishedMetadbSample getPublishedMetadbSamplebyUUID(UUID metadbSampleId) throws Exception {
+        MetadbSample sample = getMetadbSample(metadbSampleId);
+        return new PublishedMetadbSample(sample);
+    }
+
+    @Override
+    public List<PublishedMetadbSample> getPublishedMetadbSampleListByCmoId(
```
ao508, line 181: `getPublishedMetadbSampleListByCmoId` --> `getPublishedMetadbSampleListByCmoPatientId`

ao508 (changes requested):
almost there! 

there are a few minor changes left to make

ao508 (approved):

PR #472, opened 2021-11-10
Fix RequestMetadata setup for incoming requests on IGO_NEW_REQUEST topic

ao508:


divyamadala30 (approved):
Looks good. I think it's ready to be merged as long as the tests pass!

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -445,4 +448,17 @@ public void onMessage(Message msg, Object message) {
         return requestSamplesList;
     }
 
+    private RequestMetadata extractRequestMetadata(String requestMetadataJson)
```
ao508, line 451: this was refactored from the request service impl class

PR #475, opened 2021-11-15
Primary id migration 2

ao508:
TODO:

**--> this first bullet point was moved into its own scrum card for estimating** [link](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/mskcc/cmo-metadb/476)

```
- [ ] Add test case to make sure that sample service returns the same results for a given cmo patient id when (1) fetching samples by igo id and (2) fetching samples by primary Id 

Note: this case only should only work with the current mock incoming/publishing requests because the requests only contain research samples. We will also need to add a mock incoming request that we can also associate some clinical samples with. 
- mocked incoming request with a couple research samples (2 normal, 2 tumor) + 2 cmo patient ids
- these 2 cmo patient ids map to 2 dmp ids that we will have mocked clinical sample data for (1 clinical sample per cmo patient id)
- expected results when fetching by sample cmo patient ids: 1 cmo patient id => 1 research sample, 1 normal sample, 1 clinical sample
```

-------

- [ ] update methods that fetch sample by igo id to fetching sample by primary id where appropriate..

- [ ] update query for fetching sample metadata history by igo id 

```
    @Query("MATCH (sa :SampleAlias {value: $igoId, namespace: 'igoId'})-[:IS_ALIAS]->(s: Sample)"
            + "-[:HAS_METADATA]->(sm: SampleMetadata) RETURN sm")
    List<SampleMetadata> findSampleMetadataHistoryByIgoId(@Param("igoId") String igoId);
```

### DIFF
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
@@ -36,7 +36,6 @@ public class SampleMetadata implements Serializable, Comparable<SampleMetadata> 
     private List<Library> libraries;
     private String mrn;
     private String cmoPatientId;
-    @JsonProperty("primaryId")
     @JsonAlias("igoId")
     private String primaryId;
     private String investigatorSampleId;
--- a/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
+++ b/model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
@@ -34,7 +34,8 @@ public class SampleMetadata implements Serializable, Comparable<SampleMetadata> 
     private List<Library> libraries;
     private String mrn;
     private String cmoPatientId;
-    private String igoId;
+    private UUID metaDbPatientId;
+    private String primaryId;
     private String investigatorSampleId;
     private String species;
     private String sex;
@@ -61,7 +62,7 @@ public class SampleMetadata implements Serializable, Comparable<SampleMetadata> 
 
     /**
      * SampleMetadata constructor
-     * @param igoId
+     * @param primaryId
      * @param cmoInfoIgoId
      * @param cmoSampleName
      * @param sampleName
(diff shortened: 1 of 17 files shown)


persistence/src/main/java/org/mskcc/cmo/metadb/persistence/neo4j/MetadbSampleRepository.java
```diff
@@ -76,7 +76,7 @@ MetadbSample findSampleByRequestAndIgoId(@Param("reqId") String reqId,
     )
     List<SampleMetadata> findSampleMetadataListByCmoPatientId(@Param("cmoPatientId") String cmoPatientId);
 
-    @Query("MATCH (sm: SampleMetadata {igoId: $igoId})"
-            + "RETURN sm")
+    @Query("MATCH (sa :SampleAlias {value: $igoId, namespace: 'igoId'})-[:IS_ALIAS]->(s: Sample)"
+            + "-[:HAS_METADATA]->(sm: SampleMetadata) RETURN sm")
     List<SampleMetadata> findSampleMetadataHistoryByIgoId(@Param("igoId") String igoId);
```
ao508, line 81: confirm usage of this method

model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
```diff
@@ -34,7 +36,9 @@
     private List<Library> libraries;
     private String mrn;
     private String cmoPatientId;
-    private String igoId;
+    @JsonProperty("primaryId")
```
divyamadala30, line 39: I don't think this is necessary right?

service/src/test/resources/data/published_requests/outgoing_mocked_request2b_all_pairs.json
```diff
@@ -267,7 +267,7 @@
             "cmoSampleClass": "Normal", 
             "cmoSampleName": "C-PPPXX2-N001-d", 
             "collectionYear": "", 
-            "igoId": "22022_BZ_4",
```
divyamadala30, line 270: Shouldn't some of these remain as igoId to test if the json is being processed correctly and successfully?

divyamadala30 (commented):
A couple points to discuss. Aside from that, the PR looks good!

PR #492, opened 2021-11-24
SampleAliases are now added within the messageHandlingService

divyamadala30:
Only a DRAFT PR. Unit tests are still failing. 

Suspected issue: SampleAliases are not being set. When the sample is fetched, loaded and ready to saved, the sampleAliasList is still null. SampleClass is also not being set.


ao508:
case covered in #495

PR #494, opened 2021-11-29
refactoring sampleService

divyamadala30:
1. Refactored SampleService to remove any redundant code and use consistent naming convention.
2. Introduced SampleDataFactory to replace any code within the layer that populated the missing fields in MetadbSample object.


ao508:
Replaced by #495

PR #557, opened 2022-01-20
Handler for corrected cmo patient ids

divyamadala30:

- Handles corrected cmoPatientIds and makes changes accordingly

## Crossing T's and dotting I's

Please follow these checklists to help prevent any unexpected issues from being introduced by the changes in this pull request. If an item does not apply then indicate so by surrounding the line item with `~~` to strikethrough the text. See [basic writing and formatting syntax](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) for more information.

### I. Web service and data model checklist

Please follow these checks if any changes were made to any classes in the web, service, or persistence layers.

**Code checks:**
- [ ] Endpoints were tested to ensure their integrity.
- [ ] Screenshots have been provided to demonstrate changes made to the response body JSON schema and/or swagger page.
- [ ] Unit tests were updated in relation to updates to the mocked test data.

### III. Message handlers checklist:
- [ ] Changes in this PR affect the workflow of incoming messages.
- [ ] Messages are following the expected workflow when published to the topic(s) changed or introduced in this pull request.
- [ ] Unit tests were added to ensure messages are handled as expected.

If no unit tests were updated or added, then please explain why: [insert details here]

Please describe how the workflow and messaging was tested/simulated:

**Describe your testing environment:**
YET TO BE TESTED

- NATS [local, local docker, dev server, production]
- Neo4j [local, local docker, dev server, production]
- MetaDB [local, local docker, dev server, production]
- Message publishing simulation [nats cli, docker nats cli, metadb publisher tool, other (describe below)]

Other: [insert details on how messages were published or simulated for testing]

---
### General checklist:
- [ ] All requested changes and comments have been resolved.
- [ ] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)

### DIFF
--- a/service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
+++ b/service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
@@ -40,6 +40,9 @@ public class AdminMessageHandlingServiceImpl implements AdminMessageHandlingServ
 
     @Value("${request_reply.cmo_label_update_topic}")
     private String CMO_LABEL_UPDATE_REQREPLY_TOPIC;
+    
+    @Value("${mdb_admin.cmo_sample_label_update_topic}")
+    private String CMO_SAMPLE_LABEL_UPDATE;
 
     @Autowired
     private CrdbMappingService crdbMappingService;
@@ -97,19 +100,19 @@ public class AdminMessageHandlingServiceImpl implements AdminMessageHandlingServ
                     if (idCorrectionMap != null) {
                         String oldCmoPatientId = idCorrectionMap.get("oldId");
                         String newCmoPatientId = idCorrectionMap.get("newId");
+                        List<SampleMetadata> sampleMetadataList = sampleService
+                                .getSampleMetadataListByCmoPatientId(oldCmoPatientId);
                         MetadbPatient updatedPatient = patientService.updateCmoPatientId(
                                 oldCmoPatientId, newCmoPatientId);
                         
-                        List<SampleMetadata> sampleMetadataList = sampleService
-                                .getSampleMetadataListByCmoPatientId(oldCmoPatientId);
                         for (SampleMetadata sample: sampleMetadataList) {
                             sample.setCmoPatientId(newCmoPatientId);
                             LOG.info("Publishing updated sample metadata: "
                                     + sample.getPrimaryId());
                             //check if sample is research or clinical, then publish to their respective topics
                             // Topic to push updated research samples is yet to be set up
-                            // maybe IGO_SAMPLE_METADATA_UPDATE?
-                            messagingGateway.publish("TBD",
+                            // maybe CMO_SAMPLE_LABEL_UPDATE_TOPIC?
+                            messagingGateway.publish(CMO_SAMPLE_LABEL_UPDATE,
                                     mapper.writeValueAsString(sample));
                         }
                         
--- a/service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
+++ b/service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.nats.client.Message;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
```diff
@@ -83,37 +94,27 @@ public void run() {
...
-                        // what we can do here is simply log that the label change from
-                        // "old label" --> "new label" should be persisted shortly
-                        System.out.println("TODO: LOG THAT OLD LABEL IS BEING UPDATED TO "
-                                + "NEW LABEL AND SHOULD BE PERSISTED IN DB SHORTLY");
+                        String oldCmoPatientId = idCorrectionMap.get("oldId");
+                        String newCmoPatientId = idCorrectionMap.get("newId");
+                        MetadbPatient updatedPatient = patientService.updateCmoPatientId(
+                                oldCmoPatientId, newCmoPatientId);
+                        
+                        List<SampleMetadata> sampleMetadataList = sampleService
```
ao508, line 103: I still suggest moving this call so that it's ahead of the call to update the cmo patient id 

looking at how the query is set up in `MetadbSampleRepository.findAllSampleMetadataListByCmoPatientId()` - the list of sample metadata is fetched by matching on the patient alias nodes and not the sample node `cmoPatientId` property 

```
    @Query("MATCH (pa: PatientAlias {namespace: 'cmoId', value: $cmoPatientId})-[:IS_ALIAS]->"
            + "(p: Patient)-[:HAS_SAMPLE]->(s: Sample)-[:HAS_METADATA]->(sm: SampleMetadata) "
            + "MATCH (r: Request)-[:HAS_SAMPLE]->(s) SET sm.igoRequestId = r.igoRequestId "
            + "RETURN sm"
    )
    List<SampleMetadata> findAllSampleMetadataListByCmoPatientId(@Param("cmoPatientId") String cmoPatientId);
```

the way you have it set up currently would mean the CMO patient alias would have the update. So assuming the cmo patient id is updated at the `PatientAlias` node, if you query for a sample metadata list by the `oldCmoPatientId` you wouldn't get any hits. 

You can just swap the placements of these two calls

service/src/main/java/org/mskcc/cmo/metadb/service/impl/AdminMessageHandlingServiceImpl.java
```diff
@@ -83,37 +94,27 @@ public void run() {
...
+                                .getSampleMetadataListByCmoPatientId(oldCmoPatientId);
+                        for (SampleMetadata sample: sampleMetadataList) {
+                            sample.setCmoPatientId(newCmoPatientId);
+                            LOG.info("Publishing updated sample metadata: "
+                                    + sample.getPrimaryId());
+                            //check if sample is research or clinical, then publish to their respective topics
+                            // Topic to push updated research samples is yet to be set up
+                            // maybe IGO_SAMPLE_METADATA_UPDATE?
```
ao508, line 111: we'll rename the property I had defined above to indicate that it's not going to be a req-reply topic, just a regular pub-sub channel

```
CMO_LABEL_UPDATE_REQREPLY_TOPIC
```
to

```
CMO_SAMPLE_LABEL_UPDATE_TOPIC`
```
and we'll need to update the corresponding property name in `application.properties.EXAMPLE` to 

```
mdb_admin.cmo_sample_label_update_topic
```

service/src/main/java/org/mskcc/cmo/metadb/service/impl/PatientServiceImpl.java
```diff
@@ -36,5 +36,16 @@ public MetadbPatient getPatientByCmoPatientId(String cmoPatientId) {
...
         return patientRepository.findPatientIdBySample(metadbSampleId);
     }
+    
+    @Override
+    public MetadbPatient updateCmoPatientId(String oldCmoPatientId, String newCmoPatientId) {
+        if (getPatientByCmoPatientId(oldCmoPatientId) == null) {
+            return null;
+        }
+        //only update the patient if it is found, hence not null
+        patientRepository.updateCmoPatientIdInPatientNode(oldCmoPatientId, newCmoPatientId);
```
ao508, line 46: this method is returning an instance of `MetadbPatient` already isn't it? So we could just assign it to a variable and return the variable instead of calling this again  in the return statement `getPatientByCmoPatientId(newCmoPatientId)`

ao508 (changes requested):

PR #589, opened 2022-02-04
Request/Sample update message handler

divyamadala30:
# UpdateMessageHandler can now handle updates in universal schema

- Request/Sample updates are received in universal json schema
- Changes are made to adopt to the new schema 

## Crossing T's and dotting I's

Please follow these checklists to help prevent any unexpected issues from being introduced by the changes in this pull request. If an item does not apply then indicate so by surrounding the line item with `~~` to strikethrough the text. See [basic writing and formatting syntax](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) for more information.

### I. Web service and data model checklist

Please follow these checks if any changes were made to any classes in the web, service, or persistence layers.

**Data checks:**
Updates were made to the mocked incoming request data and/or mocked published request data:
- [ ] ~[cmo-metadb test data](https://github.com/mskcc/cmo-metadb/tree/master/service/src/test/resources/data)~
- [ ] ~[cmo-metadb-common test data](https://github.com/mskcc/cmo-metadb-common/tree/master/src/test/resources/data)~

**Code checks:**
- [x] Endpoints were tested to ensure their integrity.
- [x] Screenshots have been provided to demonstrate changes made to the response body JSON schema and/or swagger page.
- [ ] Unit tests were updated in relation to updates to the mocked test data.

### II. Neo4j models and database schema checklist:
- [ ] ~Neo4j persistence models were changed.~
- [ ] ~The graph database produces the expected changes to models, relationships, and/or property names. [provide screenshot of updated elements in graph db below]~

### III. Message handlers checklist:
- [x] Changes in this PR affect the workflow of incoming messages.
- [x] Messages are following the expected workflow when published to the topic(s) changed or introduced in this pull request.
- [ ] Unit tests were added to ensure messages are handled as expected.

If no unit tests were updated or added, then please explain why: [insert details here]

Please describe how the workflow and messaging was tested/simulated:

**Describe your testing environment:**

- NATS [local, **local docker**, dev server, production]
- Neo4j [local, **local docker**, dev server, production]
- MetaDB [local, **local docker**, dev server, production]
- Message publishing simulation [nats cli, docker nats cli, **metadb publisher tool**, other (describe below)]

Other: [insert details on how messages were published or simulated for testing]

---
### Screenshots

Endpoint:

- New fields introduced to `PublishedSampleMetadata`:

newfieldiscmosample

newfieldrequestId

Neo4j:

- After persisting a complete request

noupdates

- After publishing request metadata updates

requestupdates

- After publishing sample metadata updates

sampleupdates

---
### General checklist:
- [ ] All requested changes and comments have been resolved.
- [ ] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)


model/src/main/java/org/mskcc/cmo/metadb/model/MetadbRequest.java
```diff
@@ -365,8 +365,8 @@ public void updateRequestMetadataByMetadata(RequestMetadata requestMetadata)
         Map<String, Object> metadataMap =
                 mapper.readValue(requestMetadata.getRequestMetadataJson(), Map.class);
 
-        this.igoRequestId = String.valueOf(metadataMap.get("requestId"));
```
ao508, line 368: now I understand what you were referring to before - so here is where we would be processing something that's being published from LimsRest which would still have the former property names. If you want to generalize this then we can add helper methods to resolve the values for requestId and gene panel/recipe

You can refer to the helper methods added to SampleDataFactory. 

But basically the helper method should:

1. check if the map contains `requestId` and use the value if that property is present
2. if (1) fails then check for `igoRequestId` and use that value instead

model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
```diff
@@ -22,6 +23,7 @@
 import org.neo4j.ogm.annotation.typeconversion.Convert;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
divyamadala30, line 26: This is ignore any any additional properties that may come with publishedSampleJson. Used when the json is parsed to SampleMetadata in the sampleUpdateHandler

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -63,6 +65,7 @@ public PublishedMetadbSample() {}
     public PublishedMetadbSample(MetadbSample metaDbSample) throws ParseException {
         SampleMetadata latestSampleMetadata = metaDbSample.getLatestSampleMetadata();
         this.metaDbSampleId = metaDbSample.getMetaDbSampleId();
+        this.igoRequestId = latestSampleMetadata.getIgoRequestId();
```
divyamadala30, line 68: Didn't add one for isCmoSample, this manually added in the sampleServiceImpl

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/neo4j/MetadbSampleRepository.java
```diff
@@ -83,4 +83,8 @@ MetadbSample findResearchSampleByRequestAndIgoId(@Param("reqId") String reqId,
             + "RETURN sm")
     List<SampleMetadata> findSampleMetadataHistoryByNamespaceValue(
             @Param("namespace") String namespace, @Param("value") String value);
+
+    @Query("MATCH (r: Request)-[:HAS_SAMPLE]->(s: Sample {metaDbSampleId: $metaDbSampleId}) "
+            + "RETURN r.isCmoRequest")
+    String findSampleHasCmoRequest(@Param("metaDbSampleId") UUID metaDbSampleId);
```
divyamadala30, line 89: Query to check if the request of a sample is cmo or non-cmo

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbRequestService.java
```diff
@@ -19,6 +19,8 @@
     Boolean requestHasUpdates(MetadbRequest existingRequest, MetadbRequest request) throws Exception;
     Boolean requestHasMetadataUpdates(RequestMetadata existingRequestMetadata,
             RequestMetadata requestMetadata) throws Exception;
+    Boolean requestHasMetadataUpdates(String existingRequestMetadataJson,
```
divyamadala30, line 22: same as `Boolean requestHasMetadataUpdates(RequestMetadata existingRequestMetadata,
            RequestMetadata requestMetadata)` but takes different arguments

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -230,20 +230,10 @@ public void run() {
                         MetadbRequest existingRequest =
                                 requestService.getMetadbRequestById(requestMetadata.getIgoRequestId());
                         if (existingRequest == null) {
-                            LOG.info("Request does not already exist - saving to database: "
```
divyamadala30, line 233: Removed this because the request metadata we receive will not have any samples. It wouldn't make sense to persist a request with no samples

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -230,20 +230,10 @@ public void run() {
...
-                            LOG.info("Publishing new request metadata to " + CMO_REQUEST_UPDATE_TOPIC);
-                            requestService.saveRequest(request);
-                            messagingGateway.publish(request.getIgoRequestId(),
-                                    CMO_REQUEST_UPDATE_TOPIC,
-                                    mapper.writeValueAsString(
-                                          request.getRequestMetadataList()));
+                            // TODO: Log this request
                         } else if (requestService.requestHasMetadataUpdates(
-                                existingRequest.getLatestRequestMetadata(), requestMetadata)) {
+                                                mapper.writeValueAsString(existingRequest),
```
divyamadala30, line 235: ExistingRequestJson follows the universal schema like the requestMetadataUpdate we received

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -566,8 +556,10 @@ public void onMessage(Message msg, Object message) {
                 try {
                     String requestMetadataJson = mapper.readValue(
                             new String(msg.getData(), StandardCharsets.UTF_8), String.class);
+                    System.out.println("\nbefore\n" + requestMetadataJson);
```
divyamadala30, line 559: need to remove these

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
```diff
@@ -133,7 +133,10 @@ public Boolean sampleHasMetadataUpdates(SampleMetadata existingSampleMetadata,
     @Override
     public PublishedMetadbSample getPublishedMetadbSample(UUID metadbSampleId) throws Exception {
         MetadbSample sample = getMetadbSample(metadbSampleId);
-        return new PublishedMetadbSample(sample);
+        PublishedMetadbSample publishedSample = new PublishedMetadbSample(sample);
+        publishedSample.setIsCmoSample(Boolean.valueOf(sampleRepository
```
divyamadala30, line 137: This is where i add `isCmoSample` field. open to suggestions for a better name than isCmoSample

model/src/main/java/org/mskcc/cmo/metadb/model/MetadbRequest.java
```diff
@@ -385,6 +386,16 @@ public void updateRequestMetadataByMetadata(RequestMetadata requestMetadata)
...
+    public String resolveIgoRequestId(Map<String, Object> metadataMap) {
+        return ObjectUtils.firstNonNull(String.valueOf(metadataMap.get("igoRequestId")),
+                String.valueOf(metadataMap.get("requestId")));
+    }
+
+    public String resolveGenePanel(Map<String, Object> metadataMap) {
+        return ObjectUtils.firstNonNull(String.valueOf(metadataMap.get("genePanel")),
+                String.valueOf(metadataMap.get("recipe")));
+    }
+
```
ao508, line 398: I would reverse the order of the fields being checked here

model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
ao508: What additional fields did you run into? Do you have a record of which ones the deserializer complained about?

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -22,6 +22,7 @@
     @Convert(UuidStringConverter.class)
     private UUID metaDbSampleId;
     private UUID metaDbPatientId;
+    private String igoRequestId;
```
ao508, line 25: ~~note for self: still unsure about needing to add these props to the sample schema.. will think about this further.~~

edit: see proposed requested change below

ao508: [requested change]  `SampelMetadata` already contains `additonalProperties`, a hashmap of string/string. We can add _that_ to `PublishedMetadbSample` and just add this information (igoRequestId and isCmoSample) to the `additionalProperties` map.

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/neo4j/MetadbSampleRepository.java
ao508: [requested change] expanding `PublishedMetadbSample` with the `SampleMetadata.additionalProperties` will allow us to avoid having to need this query

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbRequestService.java
ao508: [requested change] something I want to dig into a bit is whether we can still treat this incoming JSON as a `RequestMetadata` object and either add a method to the request data factory that parses it as an instance of `RequestMetadata`

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
ao508: [requested change] log that the request does not already exist and so updates will not be persisted // a short blurb about how we do not persist requests that do not have any samples

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
ao508: [requested change] could we not use `existingRequest.getLatestRequestMetadata().getRequestMetadataJson()`?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestServiceImpl.java
```diff
@@ -199,8 +199,14 @@ public Boolean requestHasMetadataUpdates(RequestMetadata existingRequestMetadata
...
         String currentMetadata = mapper.writeValueAsString(requestMetadata);
+        return requestHasMetadataUpdates(existingMetadata, currentMetadata);
+    }
+
+    @Override
+    public Boolean requestHasMetadataUpdates(String existingRequestMetadataJson,
+            String requestMetadataJson) throws Exception {
         try {
-            metadbJsonComparator.isConsistent(currentMetadata, existingMetadata);
+            metadbJsonComparator.isConsistent(requestMetadataJson, existingRequestMetadataJson);
```
ao508, line 209: [requested change] Was this switch intentional? Which of these is considered the reference json and the target json?

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: [requested change] throw this into `additionalProperties` map

ao508: also can remove call to query below

service/src/main/java/org/mskcc/cmo/metadb/service/util/RequestDataFactory.java
```diff
@@ -86,8 +87,12 @@ private static RequestMetadata extractRequestMetadataFromJson(String requestMeta
         if (requestMetadataMap.containsKey("samples")) {
             requestMetadataMap.remove("samples");
         }
-        RequestMetadata requestMetadata = new RequestMetadata(
-                requestMetadataMap.get("requestId").toString(),
+        // an attempt to make this method flexible with both schemas
+        String requestId = ObjectUtils.firstNonNull(
+                String.valueOf(requestMetadataMap.get("igoRequestId")),
```
ao508, line 92: [requested change] reverse the ordering of the fields checked

ao508 (changes requested):

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/neo4j/MetadbSampleRepository.java
divyamadala30: This query is used when we're setting up new `publishedMetadbSamples` to populate `isCmoSample`

model/src/main/java/org/mskcc/cmo/metadb/model/SampleMetadata.java
divyamadala30: The alias fields

service/src/main/java/org/mskcc/cmo/metadb/service/MetadbRequestService.java
divyamadala30: That's what this method is doing, parsing the json to `requestMetadata` and then calling `requestHasMetadataUpdates(RequestMetadata existingRequestMetadata, RequestMetadata requestMetadata)`

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
divyamadala30: No because the stored `requestJson` is in the igo metadata schema not the universal schema

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestServiceImpl.java
divyamadala30: currentMetadata ->  requestMetadataJson
existingMetadata -> existingRequestMetadataJson

service/src/main/java/org/mskcc/cmo/metadb/service/util/RequestDataFactory.java
divyamadala30: 👍

model/src/main/java/org/mskcc/cmo/metadb/model/MetadbRequest.java
```diff
@@ -385,6 +386,16 @@ public void updateRequestMetadataByMetadata(RequestMetadata requestMetadata)
         addRequestMetadata(requestMetadata);
     }
 
+    public String resolveIgoRequestId(Map<String, Object> metadataMap) {
+        return ObjectUtils.firstNonNull(String.valueOf(metadataMap.get("igoRequestId")),
```
ao508, line 390: reverse the order of the fields checked here

ao508 (changes requested):
looks like there are still some changes requested that haven't been implemented

model/src/main/java/org/mskcc/cmo/metadb/model/MetadbRequest.java
```diff
@@ -385,6 +386,16 @@ public void updateRequestMetadataByMetadata(RequestMetadata requestMetadata)
         addRequestMetadata(requestMetadata);
     }
 
+    public String resolveIgoRequestId(Map<String, Object> metadataMap) {
```
ao508, line 389: just now noticed that these aren't in the RequestDataFactory class - we want to try and minimize any of this kind of logic/data extraction outside of the models

ao508: keeping this for now

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
ao508: As long as the two instances of RequestMetadata are being compared against each other then it shouldn't be problematic to determine if there are updates with the existing `requestService.requestHasMetadataUpdates(...)` method.

I still want to test this further but so far I was able to verify an `existingRequest.getLatestRequestMetadata()` against a "published message" containing only an `RequestMetadata` JSON with "updates" and the comparator was still able to handle this case as expected.

service/src/main/java/org/mskcc/cmo/metadb/service/impl/RequestServiceImpl.java
ao508: definitely the existing request metadata, meaning the one that was fetched from the database, is the one that should be on the left

ao508: this is fixed in the changes I am working on

model/src/main/java/org/mskcc/cmo/metadb/model/MetadbRequest.java
```diff
@@ -381,10 +381,14 @@ public void updateRequestMetadataByMetadata(RequestMetadata requestMetadata)
         this.libraryType = String.valueOf(metadataMap.get("libraryType"));
         this.bicAnalysis = Boolean.parseBoolean(String.valueOf(metadataMap.get("bicAnalysis")));
         this.isCmoRequest = Boolean.parseBoolean(String.valueOf(metadataMap.get("isCmoRequest")));
-        this.metaDbProject = new MetadbProject(igoRequestId.split("_")[0]);
         addRequestMetadata(requestMetadata);
     }
 
+    public String resolveGenePanel(Map<String, Object> metadataMap) {
+        return String.valueOf(ObjectUtils.firstNonNull(metadataMap.get("recipe"),
```
ao508, line 388: don't return the `String.valueOf(..)` here since it might be a null and we'd end up with a `"null"` value.

Instead look at what's returned by ObjectUtils... and then if null return an empty string

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
ao508: the request id is at the request-level, not sample level

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -220,6 +223,14 @@ public void setMetaDbPatientId(UUID metaDbPatientId) {
...
     }
 
+    public String getIgoRequestId() {
+        return igoRequestId;
+    }
+
+    public void setIgoRequestId(String igoRequestId) {
+        this.igoRequestId = igoRequestId;
+    }
+
```
ao508, line 233: remove

model/src/main/java/org/mskcc/cmo/metadb/model/web/PublishedMetadbSample.java
```diff
@@ -316,6 +327,14 @@ public void setDatasource(String datasource) {
...
     }
 
+    public Boolean getIsCmoSample() {
+        return isCmoSample;
+    }
+
+    public void setIsCmoSample(Boolean isCmoSample) {
+        this.isCmoSample = isCmoSample;
+    }
+
```
ao508, line 337: remove

service/src/test/java/org/mskcc/cmo/metadb/service/RequestServiceTest.java
```diff
@@ -162,6 +162,34 @@ public void testRequestHasMetadataUpdates() throws Exception {
         Boolean hasUpdates = requestService.requestHasUpdates(
                 origRequest, updatedRequest);
         Assertions.assertThat(hasUpdates).isEqualTo(Boolean.TRUE);
+
+
```
ao508, line 166: accidental new line added

service/src/main/java/org/mskcc/cmo/metadb/service/util/RequestDataFactory.java
```diff
@@ -86,8 +86,11 @@ private static RequestMetadata extractRequestMetadataFromJson(String requestMeta
         if (requestMetadataMap.containsKey("samples")) {
             requestMetadataMap.remove("samples");
         }
+        Object requestId = requestMetadataMap.containsKey("requestId")
```
ao508, line 89: I'm thinking that we use the same logic to resolve the genePanel above instead of the objectutils

service/src/main/java/org/mskcc/cmo/metadb/service/impl/MessageHandlingServiceImpl.java
```diff
@@ -230,18 +230,7 @@ public void run() {
...
-                            MetadbRequest request =
-                                    RequestDataFactory.buildNewRequestFromMetadata(requestMetadata);
-
-                            LOG.info("Publishing new request metadata to " + CMO_REQUEST_UPDATE_TOPIC);
-                            requestService.saveRequest(request);
-                            messagingGateway.publish(request.getIgoRequestId(),
-                                    CMO_REQUEST_UPDATE_TOPIC,
-                                    mapper.writeValueAsString(
-                                          request.getRequestMetadataList()));
+                            // TODO: Log this request
```
ao508, line 233: need to implement the actual logging here

persistence/src/main/java/org/mskcc/cmo/metadb/persistence/neo4j/MetadbSampleRepository.java
ao508: I'm still hesitant to keep this query but it's fine for now. 

can you rename to `findSampleCmoRequestStatus`

service/src/main/java/org/mskcc/cmo/metadb/service/impl/SampleServiceImpl.java
ao508: reminder: remove from published sample and expand published sample with `additionalProperties` instead, which is stored in the `SampleMetadata` node already 

then add `isCmo` to the additional properties for the published sample returned

ao508 (changes requested):

PR #595, opened 2022-02-08
Additional changes

ao508:


service/src/main/java/org/mskcc/cmo/metadb/service/util/RequestDataFactory.java
```diff
@@ -38,20 +38,6 @@ public static MetadbRequest buildNewLimsRequestFromJson(String requestJson)
...
     }
 
-    /**
-     * Method factory returns an instance of MetadbRequest built from
-     * an instance of RequestMetadata;
-     * @param requestMetadata
-     * @return MetadbRequest
-     * @throws JsonProcessingException
-     */
-    public static MetadbRequest buildNewRequestFromMetadata(RequestMetadata requestMetadata)
```
ao508, line 48: unused so this was removed

PR #630, opened 2022-03-02
CRDB service query timeout

ao508:
Now forces a timeout for a hanging query to the CRDB service.

An attempt to use @Transactional did not resolve the issue with the hanging query.


divyamadala30 (approved):

divyamadala30 (approved):

service/src/main/java/org/mskcc/cmo/metadb/service/impl/CrdbMappingServiceImpl.java
```diff
@@ -19,11 +29,41 @@ public CrdbMappingServiceImpl(CrdbRepository crdbRepository) {
...
+            public Object call() {
+                return crdbRepository.getCmoPatientIdByInputId(inputId);
+            }
+        };
+        Object result = runQueryWithForcedTimeout(task);
+        if (result != null) {
+            return result.toString();
+        }
+        return null;
+
```
divyamadala30, line 58: extra line

PR #972, opened 2023-06-06
Publish sample updates when revisable is false

ao508:
# Publish sample updates when revisable is false

- changes introduced allow support for when we migrate to allowing the dashboard to make updates to the database directly instead of waiting/polling for updates


ao508:
Note that this needs to be tested still

qu8n (approved):

PR #1015, opened 2023-08-10
Added support for mrn to cmo/dmp patient id mapping

n1zea144:
Maps a list of MRN to a list of dmp-cmo-mrn patient id triplets.

### DIFF
--- a/persistence/src/main/java/org/mskcc/smile/persistence/jpa/CrdbRepository.java
+++ b/persistence/src/main/java/org/mskcc/smile/persistence/jpa/CrdbRepository.java
@@ -1,6 +1,7 @@
 package org.mskcc.smile.persistence.jpa;
 
 import org.mskcc.smile.model.internal.CrdbMappingModel;
+import org.mskcc.smile.model.web.CrdbCrosswalkTriplet;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.query.Param;
@@ -16,4 +17,8 @@ public interface CrdbRepository extends CrudRepository<CrdbMappingModel, Long> {
             nativeQuery = true)
     Object getCmoPatientIdByInputId(@Param("inputId") String inputId);
     // another query option
+    @Query(value = "SELECT CMO_ID, DMP_ID, PT_MRN "
+           + "FROM CRDB_CMO_LOJ_DMP_MAP WHERE :inputId IN (DMP_ID, PT_MRN, CMO_ID)",
+           nativeQuery = true)
+    Object getCrdbCrosswalkTripletByInputId(@Param("inputId") String inputId);
 }
--- a/web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
+++ b/web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
@@ -1,0 +1,76 @@
+package org.mskcc.smile.web;
+
+import io.swagger.annotations.Api;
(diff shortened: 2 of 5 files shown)


web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
```diff
@@ -0,0 +1,76 @@
...
+    @Autowired
+    private CrdbMappingService crdbMappingService;
+
+    @Autowired
+    public CrosswalkController(CrdbMappingService crdbMappingService) {
+        this.crdbMappingService = crdbMappingService;
+    }
+
+    /**
+     * Returns a list of mrn-dmp-cmo patient id triplets given a list of MRNs.
```
qu8n, line 45: To confirm my understanding, technically this REST endpoint will return the triplets given not just MRNs but also DMP IDs and/or CMO Patient IDs, correct? If so, should we update the description to reflect that?

persistence/src/main/java/org/mskcc/smile/persistence/jpa/CrdbRepository.java
```diff
@@ -16,4 +17,8 @@ public interface CrdbRepository extends CrudRepository<CrdbMappingModel, Long> {
             nativeQuery = true)
     Object getCmoPatientIdByInputId(@Param("inputId") String inputId);
     // another query option
+    @Query(value = "SELECT CMO_ID, DMP_ID, PT_MRN "
+           + "FROM CRDB_CMO_DMP_MAP WHERE :inputId IN (DMP_ID, PT_MRN, CMO_ID)",
+           nativeQuery = true)
+    Object getCrdbCrosswalkTripletByInputId(@Param("inputId") String inputId);
```
qu8n, line 23: Am I understanding it correctly that the CMO-DMP-MRN mapping already exists inside CRDB, and that you just wrote a new query in order to get the triplets?

web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
qu8n, line 1: Is it possible to spin up a local Swagger REST service for testing?

qu8n (approved):
Thanks for turning around so quick, Ben. I reviewed your codes and they generally make sense. No blockers on my end, just left three questions for my own understanding.

web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
n1zea144: Yes, follow the command line instruction below to pull this PR to your local machine, then you can compile/deploy locally

n1zea144: Note, I noticed the tests failed, but thats just checkstyle.  I will fix them, but the code works.

qu8n: Good to know. Thanks!

persistence/src/main/java/org/mskcc/smile/persistence/jpa/CrdbRepository.java
n1zea144: Yes, there is a crosswalk table in CRDB with this triplet info.

web/src/main/java/org/mskcc/smile/web/CrosswalkController.java
n1zea144: Yeah, thats what I mean my mrn-dmp-cmo - thats the triplet.   The class is called CrdbCrosswalkTriplet too.

PR #1017, opened 2023-08-13
Bug fix for CORRECT_CMO_PATIENT_ID workflow

ao508:


service/src/main/java/org/mskcc/smile/service/impl/SampleServiceImpl.java
```diff
@@ -81,14 +81,29 @@ public SmileSample saveSmileSample(SmileSample
...
+            // determine whether a patient swap is required also
             if (!sample.getPatient().getSmilePatientId().equals(
                     existingSample.getPatient().getSmilePatientId())) {
                 LOG.info("Updating sample-to-patient relationship and removing connection to patient: "
                         + existingSample.getPatient().getSmilePatientId());
                 sampleRepository.removeSamplePatientRelationship(existingSample.getSmileSampleId(),
                         existingSample.getPatient().getSmilePatientId());
-                existingSample.setPatient(sample.getPatient());
+                // merge aliases from existing patient to the patient we are swapping to
+                SmilePatient existingPatient = existingSample.getPatient();
```
n1zea144, line 93: Should we remove existing patient from graph after this?

service/src/main/java/org/mskcc/smile/service/impl/CorrectCmoPatientHandlingServiceImpl.java
```diff
@@ -192,26 +193,25 @@ private void setupCorrectCmoPatientIdHandler(Gateway gateway,
...
                 try {
-                    // do not log contents of incoming message
                     String incomingDataString = mapper.readValue(
                             new String(msg.getData(), StandardCharsets.UTF_8),
                             String.class);
                     incomingDataMap = mapper.readValue(incomingDataString, Map.class);
-                } catch (JsonProcessingException e) {
-                    LOG.error("Error processing the incoming data map. "
-                            + "Refer to NATS logs for more details.");
+                    oldCmoPatientId = crdbMappingService.getCmoPatientIdByInputId(
```
n1zea144, line 203: From code, I understand these additions and what is happening in onMessage(), but I don't understand the reasoning.  For example, after attempting to set oldCmoPatientId here, it is then set on line [224](https://github.com/mskcc/smile-server/blob/95b89a6943b4697459785285a2dc5adfe6327d35/service/src/main/java/org/mskcc/smile/service/impl/CorrectCmoPatientHandlingServiceImpl.java#L224C29-L224C76) if the oldCmoPatientId was not found in CRDB, but does start with a "C-".  This is the same for newCmoPatientId.  If a CMO patient id is not found in CRDB, should we be processing it?

ao508: Yeah we decided that we would still be processing these ID's. CRDB lookup table only returns the barcode without the leading "C-"

service/src/main/java/org/mskcc/smile/service/impl/SampleServiceImpl.java
ao508: If you were to toggle the smile patient id here I think it would be the same for both of these records. If they do match then that would confirm my suspicion that the table itself is fetching and displaying duplicate results (one record for each sample) instead of results that are actually unique by patient

ao508: If the smile patient ids for these records to do not match then yeah, that would be indicative of a database issue instead

